PR #6, opened 2022-11-30
Fix build on Mac OS/libc++; allow setting build type

cataphract:


dgoffredo:
Thanks again for this contribution.

Recently, Caleb has experimented with building this library on his Macbook, but he found that another approach was needed. Then it turned out he didn't need to build this library on his Macbook after all.

I played around with a CMake-only libcurl build, but that had issues because it is unable to produce an archive (you only get `libcurl.so`).

For now, we will not support Darwin, but I will use this PR as a reference when we eventually do add support, somehow.

PR #12, opened 2022-12-14
Propagation changes in anticipation of W3C support

dgoffredo:
I'm reading through our internal RFC about adding support for W3C-style trace context propagation.

This pull request has changes in anticipation of that feature.


src/datadog/tracer_config.cpp
```diff
@@ -119,6 +122,134 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
+        |
+    tsi |  x     x     x     x     warn
+        |
+    si  |  x     x     x     x     x
+  */
+  // In each pair, the first would be overridden by the second.
+  const std::pair<env::Variable, env::Variable> questionable_combinations[] = {
+           {ts, tse}, {ts, se},  {ts, tsi}, {ts, si},
+
+                      {se, tse}, /* ok */   /* ok */
```
cgilmour, line 180: Did formatting ruin the comments here a bit?

src/datadog/tracer_config.cpp
```diff
@@ -119,6 +122,134 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
+    const auto var_name = name(var);
+    const auto var_name_override = name(var_override);
+    logger.log_error(Error{
+        Error::MULTIPLE_PROPAGATION_STYLE_ENVIRONMENT_VARIABLES,
+        warn_message(var_name, *value, var_name_override, *value_override)});
+  }
+
+  // Parse the propagation styles from the configuration and/or from the
+  // environment.
+  // Exceptions make this section simpler.
```
cgilmour, line 224: How would it look with error-handling instead of exceptions?

src/datadog/tracer_config.cpp
```diff
@@ -119,6 +122,134 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
+  try {
+    const auto global_styles = styles_from_env(env::DD_TRACE_PROPAGATION_STYLE);
```
cgilmour, line 226: By this point, I'm wondering the value here of the macro aligning variable names with environment variable names.

src/datadog/tracer_config.cpp
```diff
@@ -119,6 +122,134 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
+                 global_styles, config.extraction_styles);
+    result.injection_styles =
+        value_or(styles_from_env(env::DD_TRACE_PROPAGATION_STYLE_INJECT),
+                 styles_from_env(env::DD_PROPAGATION_STYLE_INJECT),
+                 global_styles, config.injection_styles);
+  } catch (Error &error) {
+    return std::move(error);
+  }
+
+  if (!result.extraction_styles.datadog && !result.extraction_styles.b3 &&
```
cgilmour, line 239: These expressions are tricky to reason about.

src/datadog/tracer_config.cpp
```diff
@@ -119,6 +122,134 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
+      !result.extraction_styles.none) {
+    return Error{Error::MISSING_SPAN_EXTRACTION_STYLE,
+                 "At least one extraction style must be specified."};
```
cgilmour, line 242: Could we just fall back to a default in this instance, ie: datadog extraction / injection.

cgilmour (commented):
A few comments for now

src/datadog/tracer.cpp
```diff
@@ -325,13 +340,24 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
       std::string message;
       message += "B3 extracted different data than did ";
       message += extracted_by;
-      // TODO: diagnose difference
+      message += ".  B3 extracted ";
```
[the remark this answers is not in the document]
dgoffredo: Yes, I will remove this feature in an upcoming edit to this PR.

The new way will be that the list of styles represents a preference order.

src/datadog/tracer_config.cpp
dgoffredo: It would be a lot longer.

src/datadog/tracer_config.cpp
dgoffredo: The benefit is that there's one way to add an environment variable: by adding an enum value.  The list of environment variables consumed by nginx is then automatically up-to-date.

src/datadog/tracer_config.cpp
dgoffredo: A conjunction of three boolean expressions is tricky to reason about?

I added another one here (`none`) and will add at least another in an upcoming pull request (`w3c` or whatever it's called). Maybe we're getting to the point where the `PropagationStyles` type should have set-like or bitset-like semantics instead of being a `struct` of `bool`s.  I'll consider other options; it might be nice to have the equivalent of:
```c++
if (styles.all_unset()) {
```

src/datadog/tracer_config.cpp
dgoffredo: True. The thing is, the only way to get this error is if the user explicitly said "the list of styles is empty."

If empty is a suitable default, then why have an explicit "none" style, especially if "none" only makes sense by itself? Better to just have "no styles" mean "no propagation."

Alas, "none" is in our spec.

Since the easiest way to get the default behavior is to not configure the styles, I'd rather have this error than add another way to get the default behavior.

dgoffredo:
Thanks for the review.  I'll continue adding to this PR today.

src/datadog/tracer.cpp
```diff
@@ -174,6 +174,21 @@ bool operator!=(const ExtractedData& left, const ExtractedData& right) {
          left.sampling_priority != right.sampling_priority;
 }
 
+nlohmann::json to_json(const ExtractedData& data) {
+  auto result = nlohmann::json::object({});
+#define ADD_FIELD(FIELD)          \
```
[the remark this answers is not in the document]
dgoffredo: Yeah I might remove this. But "don't repeat yourself" :grimacing:

src/datadog/tracer_config.cpp
dgoffredo: They're meant to reflect the diagram commented above. I used vim for fixed with fonts :)

Maybe something was lost in translation, I'll check...

PR #13, opened 2022-12-20
W3C-style trace context propagation

dgoffredo:
This implements [W3C-style][1] trace context propagation using the "traceparent" and "tracestate" headers. It is based on a internal design document.

It's an extension of work done previously in https://github.com/DataDog/dd-trace-cpp/pull/12.

In addition to adding the `PropagationStyle::W3C`, this work also resulted in the following changes:

- `struct ExtractedData` was factored out into its own `extracted_data.{h,cpp}`.
- The `hex` function was factored out into its own `hex.{h,cpp}`.
- `bin/check` was added, and `bin/test` modified, for minor development convenience.
- The representation of trace tags is now `vector<pair<string, string>>` instead of the previous `unordered_map<string, string>`. This made unit tests simpler and more honest, and the preservation of order is a nice property to have, likely without any performance penalty.
- Trailing whitespace was removed from the code.

[1]: https://www.w3.org/TR/trace-context/


src/datadog/hex.h
```diff
@@ -0,0 +1,31 @@
...
+#include <utility>
+
+namespace datadog {
+namespace tracing {
+
+// Return the specified `value` formatted as a lower-case hexadecimal string
+// without any leading zeroes.
+template <typename Integer>
+std::string hex(Integer value) {
+  // 4 bits per hex digit char, and then +1 char for possible minus sign
```
cgilmour, line 19: Negative hex values shouldn't be a thing here, since it's intended to be an encoding of a bit pattern (more like a hexdump) instead of a representation of a numeric value.

src/datadog/trace_segment.cpp
```diff
@@ -1,40 +1,56 @@
...
+    std::string message;
+    message +=
+        "Serialized x-datadog-tags header value is too large.  The configured "
+        "maximum size is ";
+    message += std::to_string(tags_header_max_size);
+    message += " bytes, but the encoded value is ";
+    message += std::to_string(encoded_trace_tags.size());
+    message += " bytes.";
+    logger.log_error(message);
+    local_root_tags[tags::internal::propagation_error] = "inject_max_size";
```
cgilmour, line 50: Why on local-root instead of the span specifically?
What if the same span is propagated more than once?
What if multiple child spans of the local-root experienced the same error?

src/datadog/trace_segment.cpp
```diff
@@ -48,8 +64,11 @@ TraceSegment::TraceSegment(
     const std::vector<PropagationStyle>& injection_styles,
     const Optional<std::string>& hostname, Optional<std::string> origin,
     std::size_t tags_header_max_size,
-    std::unordered_map<std::string, std::string> trace_tags,
+    std::vector<std::pair<std::string, std::string>> trace_tags,
```
cgilmour, line 67: That's a lot of parameters for a constructor. Just sayin.

src/datadog/trace_segment.cpp
```diff
@@ -202,50 +225,45 @@ void TraceSegment::update_decision_maker_trace_tag() {
 
   assert(sampling_decision_);
 
+  const auto found = std::find_if(
```
cgilmour, line 228: If this code gets refactored in the future, some special care will be required for the points where `found` or `value` become invalid. A comment for future-self at those points might be a good idea.

src/datadog/w3c_propagation.cpp
```diff
@@ -0,0 +1,396 @@
...
+      "([0-9a-f]{2})"  // hex "trace-flags" (match group 5)
+      "(?:$|-.*)";     // either the end, or a hyphen preceding further fields
+  static const std::regex regex{pattern};
+
+  std::match_results<StringView::iterator> match;
+  if (!std::regex_match(traceparent.begin(), traceparent.end(), match, regex)) {
+    return "malformed_traceparent";
+  }
+
+  assert(match.ready());
```
cgilmour, line 64: Do you want any of these asserts checked at runtime, or not so much?

src/datadog/w3c_propagation.cpp
```diff
@@ -0,0 +1,396 @@
...
+namespace {
+
+// Return a predicate that returns whether its `char` argument is any of the
+// following:
+//
+// - outside of the ASCII inclusive range `[lowest_ascii, highest_ascii]`
+// - equal to one of the `disallowed_characters`.
+//
+// `verboten` is used with `std::replace_if` to sanitize field values within the
+// tracestate header.
```
cgilmour, line 25: The entire header or just our part of it?

src/datadog/w3c_propagation.cpp
```diff
@@ -0,0 +1,396 @@
...
+    // Replace them with tildes ("~").
+    std::replace(result.end() - value.size(), result.end(), '=', '~');
+  }
+
+  if (additional_datadog_w3c_tracestate) {
+    result += ';';
+    result += *additional_datadog_w3c_tracestate;
+  }
+
+  const std::size_t max_size = 256;
```
cgilmour, line 366: Is this an arbitrary size or is it defined somewhere?

test/mocks/loggers.cpp
```diff
@@ -1 +1,17 @@
...
+#include <ostream>
+
+std::ostream& operator<<(std::ostream& stream,
+                         const std::vector<MockLogger::Entry>& entries) {
+  stream << "<BEGIN " << entries.size() << " LOG ENTRIES>";
+  for (std::size_t i = 0; i < entries.size(); ++i) {
+    const auto& entry = entries[i];
+    const auto kind_name =
+        entry.kind == MockLogger::Entry::ERROR ? "ERROR" : "STARTUP";
+    stream << '\n' << (i + 1) << ". " << kind_name << ": ";
```
cgilmour, line 13: prefixing `\n` instead of suffixing it seems a bit odd, is there a corner-case driving that?

cgilmour (approved):
I'm marking this as approved for now, with some questions in the comments.
Some extra time would be needed to dig into the implementation details / cross-check with opentelemtry specs and other implementations.

Especially with regard to extraction, I think as other implementations are being completed, it'd be good to have some interop tests with other tracers developed by datadog and also external implementations.

In that circumstance, what's the overall impact on tracing if we received this header but didn't perfectly comprehend it?

src/datadog/hex.h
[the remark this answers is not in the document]
dgoffredo: This function supports any integral type, so I had to accommodate the minus sign.

In a more recent PR, I constrained the type to be unsigned and added a "padded" variant of the function: https://github.com/DataDog/dd-trace-cpp/pull/17/files#diff-666374b8256a8b91fa9a3a030a39c3f75cfafe26d044c6c9c7d9e39db6c895f8R32-R51

Would you like me to backport that change on this PR?

src/datadog/trace_segment.cpp
dgoffredo: `tags::internal::propagation_error` is specified in the horizontal propagation RFC. There it's defined to be set on the local root span, and implementations are allowed to either keep the first error or to keep the last error. I chose the latter.

src/datadog/trace_segment.cpp
dgoffredo: Yes, I could create a `struct` to hold them, but I decided not to. What do you think?

src/datadog/trace_segment.cpp
dgoffredo: I'll add comments.

src/datadog/w3c_propagation.cpp
dgoffredo: Just our part of it.

src/datadog/w3c_propagation.cpp
dgoffredo: I want them checked at runtime in debug builds, that's why I use the `assert` macro.

src/datadog/w3c_propagation.cpp
dgoffredo: It's defined in our W3C propagation RFC. It's not configurable because it is derived from the W3C definition itself.

test/mocks/loggers.cpp
dgoffredo: I don't see the difference. I'll suffix them instead.

dgoffredo: I think what I was going for is "newlines only if there are entries," but this code doesn't achieve that. I'll fix it.

PR #17, opened 2023-01-11
128-bit trace IDs

dgoffredo:
This revision introduces 128-bit trace ID generation and propagation for all propagation styles.

It's based on an internal design document.

128-bit generation is disabled by default, and can be enabled via either `bool TracerConfig::trace_id_128_bit` or `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`.


src/datadog/tracer_config.h
```diff
@@ -87,7 +87,6 @@ struct TracerConfig {
 
   // `logger` specifies how the tracer will issue diagnostic messages.  If
   // `logger` is null, then it defaults to a logger that inserts into
-  // `std::cerr`.
```
dgoffredo, line 90: put it back...

src/datadog/expected.h
```diff
@@ -17,7 +17,6 @@
 //       auto maybe_int = parse_integer("the answer");
 //       // using the `if_error` method
 //       if (auto *error = maybe_int.if_error()) {
-//         std::cerr << "parse_integer returned error: " << *error << '\n';
```
dgoffredo, line 20: put it back...

src/datadog/expected.h
```diff
@@ -26,7 +25,6 @@
 //       maybe_int = parse_integer("one hundred twenty-three");
 //       // use the `error` method.
 //       if (!maybe_int) {
-//         std::cerr << "parse_integer returned error: " << maybe_int.error() <<
```
dgoffredo, line 29: put it back...

test/test_tracer.cpp
```diff
@@ -507,6 +541,12 @@ TEST_CASE("span extraction") {
     REQUIRE(result.error().code == Error::NO_SPAN_TO_EXTRACT);
   }
 
+  SECTION("sanity") {
```
dgoffredo, line 544: Get rid of this...

src/datadog/trace_id.h
```diff
@@ -0,0 +1,53 @@
...
+
+  // Create a trace ID whose lower 64 bits are the specified `low` and whose
+  // higher 64 bits are the specified `high`.
+  TraceID(std::uint64_t low, std::uint64_t high);
+
+  // Return a 32 character lower-case hexadecimal representation of this trace
+  // ID, padded with zeroes on the left.
+  std::string hex_padded() const;
+
+  // Return either a decimal or a "0x"-prefixed hexadecimal representation of
```
cgilmour, line 36: Not sure I like this, a straight token is easier to lug around (copy-paste) if it's being used for error diagnostics.

cgilmour (approved):
Approving, with some comments for consideration.

src/datadog/trace_id.cpp
```diff
@@ -0,0 +1,95 @@
...
+bool operator==(TraceID left, TraceID right) {
+  return left.low == right.low && left.high == right.high;
+}
+
+bool operator!=(TraceID left, TraceID right) {
+  return left.low != right.low || left.high != right.high;
+}
+
+bool operator==(TraceID left, std::uint64_t right) {
+  return left == TraceID{right};
```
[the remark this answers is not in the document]
dgoffredo: `left.high == 0 && left.low == right`, as I had it before. Then I saw this as better. I see your point, though.

Think I'll keep it as is.

src/datadog/trace_sampler.cpp
```diff
@@ -40,7 +40,7 @@ SamplingDecision TraceSampler::decide(const SpanData& span) {
     decision.limiter_max_per_second = limiter_max_per_second_;
     decision.configured_rate = rule.sample_rate;
     const std::uint64_t threshold = max_id_from_rate(rule.sample_rate);
-    if (knuth_hash(span.trace_id) < threshold) {
+    if (knuth_hash(span.trace_id.low) < threshold) {
```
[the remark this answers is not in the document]
dgoffredo: Yes... maybe in `sampling_util`, where `knuth_hash` is defined. Will do.

src/datadog/tracer_config.cpp
```diff
@@ -360,6 +360,13 @@ Expected<FinalizedTracerConfig> finalize_config(const TracerConfig &config) {
   result.report_hostname = config.report_hostname;
   result.tags_header_size = config.tags_header_size;
 
+  if (auto enabled_env =
+          lookup(environment::DD_TRACE_128_BIT_TRACEID_ENABLED)) {
+    result.trace_id_128_bit = !falsy(*enabled_env);
```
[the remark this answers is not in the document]
dgoffredo: Yes, that's the desired behavior.

As we discussed in our standup yesterday, the falsiness behavior of this library could use revisiting. For now, though, I think that
```
export DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=''
```
should enable the feature, if only because of the word "enabled" in there. Reminds me of the tension in C preprocess macros between `#ifdef` and `#if`.

src/datadog/tracer_config.h
```diff
@@ -97,6 +97,16 @@ struct TracerConfig {
...
+
+  // `trace_id_128_bit` indicates whether the tracer will generate and interpret
+  // the higher 64 bits of trace IDs. If true, the tracer will generate 128-bit
+  // trace IDs, extract them from incoming trace context, inject them into
+  // outgoing trace context, and send them to the collector. If false, then the
+  // tracer will generate 64-bit trace IDs, and extract only the lower 64 bits
+  // of trace IDs from incoming trace context.
+  // `trace_id_128_bit` is overridden by the `DD_TRACE_128_BIT_TRACEID_ENABLED`
+  // environment variable.
+  bool trace_id_128_bit = false;
```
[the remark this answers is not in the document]
dgoffredo: It would mean we always send the `_dd.p.tid` trace tag to the Agent and to forward services. The data cost is marginal, and I wasn't able to measure the latency cost. Still, the spec.

PR #21, opened 2023-01-20
don't let DatadogAgent::flush run concurrently

dgoffredo:
Unlikely, but possible.

### DIFF
--- a/src/datadog/datadog_agent.cpp
+++ b/src/datadog/datadog_agent.cpp
@@ -132,32 +132,12 @@ std::variant<CollectorResponse, std::string> parse_agent_traces_response(
   return message;
 }
 
-class OneAtATime {
-  std::atomic<bool>& is_busy_;
-  bool was_busy_;
-
- public:
-  explicit OneAtATime(std::atomic<bool>& is_busy)
-      : is_busy_(is_busy), was_busy_(false) {
-    is_busy.compare_exchange_weak(was_busy_, true);
-  }
-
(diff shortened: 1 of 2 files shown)


src/datadog/datadog_agent.cpp
```diff
@@ -184,6 +204,12 @@ nlohmann::json DatadogAgent::config_json() const {
 }
 
 void DatadogAgent::flush() {
+  OneAtATime guard{flushing_};
+  if (guard.busy()) {
```
cgilmour, line 208: How can this be tested in a repeatable / reliable way without introducing additional state?
Also it has the smell of "double checked locking" edge cases, which are difficult to both prove and disprove.

cgilmour (commented):
Just a comment so we can discuss in more detail sometime

src/datadog/datadog_agent.cpp
dgoffredo: I could test it by exposing the `atomic<bool>` to a test for fiddling.

Alternatively, (and I prefer it, for some reason), I could test it "for real" by giving `DatadogAgent` an `HTTPClient` that I've mocked to block on command. Then I can play games with a mocked `EventScheduler` and the `HTTPClient` to create any interleaving of operations I want. Might need to spawn a couple threads.

By "double checked locking" edge cases, do you mean footguns in the atomic operations, or do you mean an interaction between the atomic and the mutex below it?

PR #22, opened 2023-03-15
more Span accessors, for use in Envoy's unit tests

dgoffredo:
Envoy is being modified to use `dd-trace-cpp` in place of `dd-opentracing-cpp`.

The lower level Datadog tracing components within Envoy can use a mock `Collector` to inspect span data. However, the higher level components do not allow for the injection of a `Collector` (part of their role is to specify the `Collector`). They have access to the `datadog::tracing::Span`, though, so let's make that object easier to inspect.


cgilmour (approved):
Looks OK to me.

PR #23, opened 2023-03-17
add missing overloads to TraceID's operator== and operator!=

dgoffredo:
`trace_id == some_int` was defined, but not `some_int == trace_id`.


dgoffredo:
clint

PR #25, opened 2023-03-30
cmake-native libcurl build

dgoffredo:
@cgilmour Maybe this will help build on Darwin. I don't know yet because I don't have a Mac.

Also, I've only compiled and run the unit tests. Will need to play with coverage reporting, sanitizers, fuzzing, etc.

Tomorrow.


dgoffredo:
This produces a shared library build of libcurl, where I'd prefer a static build.

I might revisit this in the future. For now, there are workarounds for building on Darwin, which isn't yet strictly required.

PR #26, opened 2023-03-30
Allow infinity and NaN to be parsed.

dgoffredo:
When building with libc++, as opposed to libstdc++, `operator>>(istream&, double&)` accepts "inf" for infinity and "nan" for NaN. Account for this in some unit tests.

This also modifies the build to use libc++ whenever clang++ is the compiler, instead of the ubuntu default of libstdc++. This required some tweaks to the CI due to a bug in how ubuntu packages clang and libc++.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,8 @@ jobs:
     resource_class: xlarge
     environment:
       MAKE_JOB_COUNT: 8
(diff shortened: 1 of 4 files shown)


cgilmour (approved):
It's one way to do it. Approving.

cgilmour:
I'd have expected the llvm toolchain that's already in CI would use libc++ by default. (it has a special option to use libstdc++).

PR #29, opened 2023-04-27
Add Benchmarking Platform support

ddyurchenko:
Adds a boilerplate to run benchmarks and report in Benchmarking Platform format.


dgoffredo:
Thanks for setting this up! Work will continue in <https://github.com/DataDog/dd-trace-cpp/pull/32> instead of here.

PR #33, opened 2023-05-10
allow for multiple examples

dgoffredo:
I'm going to be adding an example setup that has a minimal HTTP server (using a header-only C++ library) and an upstream Python server that talks to a database (SQLite).

Before that, I need to rearrange the example(s) directory, CMake build, wrapper script, and documentation. This way, when I add the new example, this boilerplate will already be done.


cgilmour (approved):
Marking as approved, but this seems more of a half-way measure in setting this up for multiple examples with just one example present.

The cmake option should probably have a top-level BUILD_EXAMPLES on-off, and if you really want specific ones for each example, that's fine.
The target within the hasher example should be renamed instead of still being just `example`, so that the next one has a pattern to follow.

There'll probably be a similar scale of changes to existing bits when the actual second example is added.

dgoffredo:
Thanks for the review!

> Marking as approved, but this seems more of a half-way measure in setting this up for multiple examples with just one example present.

The goal was to reduce the size of the diff in the subsequent pull request, but I forgot to set its base branch correctly :)

> The cmake option should probably have a top-level BUILD_EXAMPLES on-off, and if you really want specific ones for each example, that's fine.

Yeah, I started with one-option-builds all, but then it felt too much like I was building a bunch of unrelated things. I figure a user will try examples one at a time, if at all.

> The target within the hasher example should be renamed instead of still being just example, so that the next one has a pattern to follow.

Good idea, will do in a subsequent PR.

> There'll probably be a similar scale of changes to existing bits when the actual second example is added.

Yep!

Issue #31, opened 2023-05-07
curl.cpp leaks CURL handles


dgoffredo:
`curl_easy_cleanup` is not called on all `CURL*` objects.

`curl_easy_cleanup` is called (via `CurlLibrary::easy_cleanup`) when there's a failure to set up the handle (in `post`), and when the handle is completed or failed (in `handle_message`). It is _not_ called for handles that have not completed by the time we're shutting down (at the end of `run`).

Add a call to `curl_.easy_cleanup(handle)` after this line: https://github.com/DataDog/dd-trace-cpp/blob/9ac719cc1d4b6ea0a06313344c2ee432da8c9da5/src/datadog/curl.cpp#L508

PR #36, opened 2023-05-18
prevent leaking of CURL handles

dgoffredo:
See <https://github.com/DataDog/dd-trace-cpp/issues/31>.

In addition to the recommended change, this revision also adds unit tests that verify that a CURL handle, once created, is destroyed.

### DIFF
--- a/test/test_curl.cpp
+++ b/test/test_curl.cpp
@@ -9,103 +9,131 @@
 #include <chrono>
 #include <exception>
 #include <system_error>
+#include <unordered_set>
 
 #include "mocks/loggers.h"
 #include "test.h"
 
 using namespace datadog::tracing;
 
+class SingleRequestMockCurlLibrary : public CurlLibrary {
+ public:
+  void *user_data_on_header_ = nullptr;
+  HeaderCallback on_header_ = nullptr;
+  void *user_data_on_write_ = nullptr;
+  WriteCallback on_write_ = nullptr;
+  CURL *added_handle_ = nullptr;
+  CURLMsg message_;
+  // Since `SingleRequestMockCurlLibrary` supports at most one request,
+  // `created_handles_` and `destroyed_handles_` will have size zero or one.
+  std::unordered_multiset<CURL *> created_handles_;
+  std::unordered_multiset<CURL *> destroyed_handles_;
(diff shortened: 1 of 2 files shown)


test/test_curl.cpp
```diff
@@ -9,103 +9,131 @@
...
-    CURLcode easy_setopt_headerfunction(CURL *,
-                                        HeaderCallback on_header) override {
-      on_header_ = on_header;
-      return CURLE_OK;
-    }
-    CURLcode easy_setopt_writedata(CURL *, void *data) override {
-      user_data_on_write_ = data;
-      return CURLE_OK;
-    }
+class SingleRequestMockCurlLibrary : public CurlLibrary {
```
dgoffredo, line 19: This code is moved and renamed from the `class MockCurlLibrary` that used to be in the `TEST_CASE("parse response headers and body")`. It's used there and in the new test case.

I also added data members `created_handles_`, `destroyed_handles_`, `message_result_`, and `delay_message_`.

test/test_curl.cpp
```diff
@@ -344,5 +373,65 @@ TEST_CASE("setopt failures") {
...
+    const auto result = client->post(url, ignore, "whatever", ignore, ignore);
+
+    REQUIRE(result);
+    // Destroy the `Curl` object.
+    client.reset();
+  }
+
+  // Here are the checks relevant to this test.
+  REQUIRE(library.created_handles_.size() == 1);
+  REQUIRE(library.created_handles_ == library.destroyed_handles_);
```
dgoffredo, line 433: The `SECTION`s above are all setup. The post-condition, here, is always the same: one handle created and one handle destroyed.

dgoffredo:
I find that "hide whitespace" in the diff gearbox menu improves the diff.

cgilmour (approved):
Seems OK, marking as approved.
Are there any consequences to doing this with a request in-flight?

dgoffredo:
Thanks!

> Are there any consequences to doing this with a request in-flight?

Nope, removing the easy handle from the multi-handle (event loop) unregisters it from anything that would drive its installed callbacks. Then cleaning up the easy handle frees resources.

PR #39, opened 2023-06-13
suppress compiler errors stemming from bugged warning

dgoffredo:
I previously made this change to [nginx-datadog][1], which builds dd-trace-cpp as part of its build. I didn't make the corresponding change in dd-trace-cpp itself because the build image that we use for CI isn't based on Alpine, which packages a newer GCC that has this [bug][2].

[1]: https://github.com/DataDog/nginx-datadog
[2]: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108088

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,10 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# This warning has a false positive. See
+# <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108088>.
+add_compile_options(-Wno-error=free-nonheap-object)
+
 # If we're building with clang, then use the libc++ version of the standard
 # library instead of libstdc++. Better coverage of build configurations.
 #


CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# This warning has a false positive. See
```
cgilmour, line 20: It'd be nice to have an easy / automated way to know when this can be removed again.
But it's not worth putting additional effort into it.

cgilmour (approved):
Approved, and it'll be good when the hack/workaround can be removed in the future.

CMakeLists.txt
dgoffredo: CMake has a feature that allows the output of a program to be used as the value of a variable.

I wrote a `bin/spurious-warning` script that uses `g++` and `grep` to demonstrate whether the spurious warning is produced. But, I had trouble coming up with a minimum reproducing program. In particular, the snippet in the linked bugzilla report didn't trigger the warning in GCC 12 (which is what recent Alpines use).

You're right, it's not worth putting additional effort into it.

dgoffredo:
Thanks for the review!

PR #41, opened 2023-07-06
pass Datadog's internal shared tests for 128-bit trace IDs

dgoffredo:
These are the changes necessary for dd-trace-cpp to pass applicable test cases in `tests/parametric/test_128_bit_traceids.py` in Datadog's internal shared tests.

See the commit messages for the correspondence with shared tests.

### DIFF
--- a/src/datadog/id_generator.cpp
+++ b/src/datadog/id_generator.cpp
@@ -1,6 +1,7 @@
 #include "id_generator.h"
 
 #include <bitset>
+#include <chrono>
 
 #include "random.h"
 
@@ -15,11 +16,17 @@ class DefaultIDGenerator : public IDGenerator {
   explicit DefaultIDGenerator(bool trace_id_128_bit)
       : trace_id_128_bit_(trace_id_128_bit) {}
 
-  TraceID trace_id() const override {
+  TraceID trace_id(const TimePoint& start) const override {
     TraceID result;
     result.low = random_uint64();
     if (trace_id_128_bit_) {
-      result.high = random_uint64();
+      // Highest 32 bits contain a unix timestamp (the trace start time).
+      const auto since_epoch = start.wall.time_since_epoch();
+      const auto seconds =
+          std::chrono::duration_cast<std::chrono::seconds>(since_epoch).count();
+      // The farthest we'll go back is the unix epoch.
+      const std::uint64_t unsigned_seconds = seconds < 0 ? 0 : seconds;
+      result.high = unsigned_seconds << 32;
     } else {
       // In 64-bit mode, zero the most significant bit for compatibility with
       // older tracers that can't accept values above
--- a/test/test_tracer.cpp
+++ b/test/test_tracer.cpp
@@ -1083,7 +1083,20 @@ TEST_CASE("128-bit trace IDs") {
   REQUIRE(*high == trace_id.high);
 }
 
-TEST_CASE("_dd.p.tid inconsistent with trace ID results in error tag") {
+TEST_CASE(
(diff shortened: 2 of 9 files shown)


test/test_tracer.cpp
```diff
@@ -966,32 +1014,22 @@ TEST_CASE("create 128-bit trace IDs") {
...
-    CAPTURE(logger->entries);
-    REQUIRE(logger->error_count() == 0);
-    REQUIRE(collector->span_count() == 1);
-    const auto& span = collector->first_span();
-    const auto found = span.tags.find(tags::internal::trace_id_high);
-    REQUIRE(found != span.tags.end());
-    const auto high = parse_uint64(found->second, 16);
-    REQUIRE(high);
-    REQUIRE(*high == generated_id.high);
+    const std::uint64_t expected = std::uint64_t(flash_crash) << 32;
```
cgilmour, line 1027: I'd probably compare this with a raw expected value instead of a computed one.

cgilmour (approved):
Marking as approved.
Maybe some more checks and tests around zero-values will be needed for other system-tests.
Do all the related tests with valid propagation styles now pass? Any of them worth adding to this repo's unit tests?

src/datadog/id_generator.cpp
```diff
@@ -15,11 +16,17 @@ class DefaultIDGenerator : public IDGenerator {
...
     TraceID result;
     result.low = random_uint64();
     if (trace_id_128_bit_) {
-      result.high = random_uint64();
+      // Highest 32 bits contain a unix timestamp (the trace start time).
+      const auto since_epoch = start.wall.time_since_epoch();
+      const auto seconds =
+          std::chrono::duration_cast<std::chrono::seconds>(since_epoch).count();
+      // The farthest we'll go back is the unix epoch.
+      const std::uint64_t unsigned_seconds = seconds < 0 ? 0 : seconds;
```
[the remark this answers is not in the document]
dgoffredo: I agree it's not going to happen. But why 1 and not 0?

My understanding of unix time is that is the number of seconds elapsed since 00:00:00 UTC on 1 January 1970.

Hm, wait...
```console
david@ein:~/Downloads$ python
Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import datetime     
>>> datetime.datetime.fromtimestamp(0)
datetime.datetime(1969, 12, 31, 19, 0)
```
Ah, that's a time zone thing.

Anyway, the reason I have a cutoff here is to prevent the rollover to 18446744073709551615 at -1 (which is not a valid unix timestamp, but would correspond to a date over 584 trillion years in the future -- build software to last!).

test/test_tracer.cpp
dgoffredo: That's a fine line and I see your point. The test is kind of circular if it just does what the implementation does.

But if I were to include the raw expected value in the test, I'd obtain it by calculating `1273171513 << 32` (which is `5468230010533838848`). At that point we'd be testing that `operator<<(uint64_t, int)` behaves the same way for all build targets, which is meh.

dgoffredo:
Thanks for the review, Caleb.

> Maybe some more checks and tests around zero-values will be needed for other system-tests.

Maybe so. I'm starting those today.

> Do all the related tests with valid propagation styles now pass?

All tests in `tests/parametric/test_128_bit_traceids.py` pass, except those involving the single-header B3 propagation style, which we don't (yet) support.

> Any of them worth adding to this repo's unit tests?

I added unit tests wherever the library's behavior departed from that expected by the shared tests. The consistent behavior is already covered by existing unit tests.

PR #43, opened 2023-07-10
W3C propagation internal shared tests

dgoffredo:
This revision modifies the library to satisfy most of Datadog's internal shared integration tests.

Some of the tested behavior we're exempt from, or requires further discussion with the authors of the tests.

The behavior I did change is:

- Extracting a trace ID with the value zero is an error.
- When extracting the `traceparent` header for the W3C style, trailing hyphen-separated fields are tolerated only if the version field is not `00`. The idea is that the format for version `00` is known completely, and does not have trailing fields. Future versions (e.g. `01`, `02`...) may have trailing fields.
- When tunneling `x-datadog-origin` inside of the `tracestate` header's `dd` entry, equal signs must be preserved, because equal signs are allowed in `x-datadog-origin`. However, the equal sign is a special character in `tracestate`. So, instead, tildes are disallowed, so that equal signs can be converted to and from tildes.
- The default propagation style used to be "prefer W3C, otherwise Datadog." This caused issues when certain hosting environments would automatically generate W3C headers for inbound HTTP requests. As a workaround, the default propagation style is being reverted to "just Datadog." Customer can still set any propagation styles they like via the relevant `TracerConfig` fields or the `DD_TRACE_PROPAGATION_STYLE*` environment variables.


test/test_tracer_config.cpp
```diff
@@ -1003,12 +1003,12 @@ TEST_CASE("TracerConfig propagation styles") {
   TracerConfig config;
   config.defaults.service = "testsvc";
 
-  SECTION("default styles are [W3C, Datadog]") {
+  SECTION("default style is Datadog") {
     auto finalized = finalize_config(config);
     REQUIRE(finalized);
 
     const std::vector<PropagationStyle> expected_styles = {
```
cgilmour, line 1010: Probably don't need a vector of 1 item now

src/datadog/w3c_propagation.cpp
```diff
@@ -190,6 +196,7 @@ void parse_datadog_tracestate(ExtractedData& result, StringView datadog_value) {
     const auto value = range(kv_separator + 1, pair_end);
     if (key == "o") {
       result.origin = std::string{value};
+      std::replace(result.origin->begin(), result.origin->end(), '~', '=');
```
cgilmour, line 199: You'll appreciate a comment here a few months from now, to remind why this was done.
The corresponding tests don't really point out the what and why of this

cgilmour (commented):
A couple of comments for now

src/datadog/w3c_propagation.cpp
```diff
@@ -62,18 +63,23 @@ Optional<std::string> extract_traceparent(ExtractedData& result,
...
+  assert(match.size() == 5 + 1);
 
   const auto to_string_view = [](const auto& submatch) {
     assert(submatch.first <= submatch.second);
     return StringView{submatch.first,
                       std::size_t(submatch.second - submatch.first)};
   };
 
-  if (to_string_view(match[1]) == "ff") {
+  const auto version = to_string_view(match[1]);
```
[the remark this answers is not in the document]
dgoffredo: The only value that is not supported is `ff`.

I added the case with `00` because only with `00` do we reject trailing hyphen-separated fields. If it were, say, `02`, we tolerate them. This is per the shared tests.

src/datadog/w3c_propagation.cpp
dgoffredo: Will do.

cgilmour (approved):
Approving. This has the expected number of changes based on review comments

PR #45, opened 2023-08-01
remove .cpp files that do not contain definitions

dgoffredo:
This project isn't large enough to benefit from the "every header has a matching source file" rule, which itself is uncommon.


dgoffredo:
The benchmark failure is an issue with the benchmarking framework's internal report analysis, which I cannot run locally. I'll notify its maintainer.

In the mean time, I'll merge this. @cgilmour has reviewed it.

Issue #46, opened 2023-08-11
typo in documentation


dgoffredo:
https://github.com/DataDog/dd-trace-cpp/blob/72ecd8fbc13c1a2f70a56db4e3bf22e3383a3e65/src/datadog/trace_sampler.h#L7 should say "`TracerConfig::trace_sampler`", not "`TracerConfig::span_sampler`".

Issue #40, opened 2023-06-20
spelling mistake


dgoffredo:
https://github.com/DataDog/dd-trace-cpp/blob/0536a1653c6d689ab5f29d9b9defb6532e4b1b5b/src/datadog/span.h#L163

It's "affect," not "effect."

Issue #37, opened 2023-05-25
nullptr -> nullopt


dgoffredo:
https://github.com/DataDog/dd-trace-cpp/blob/338a87ed83a2fd5a89cc136c544368ddab0f20b0/src/datadog/environment.h#L75

`datadog::tracing::nullopt`, not `std::nullptr`.

PR #47, opened 2023-08-15
fix some comments

dgoffredo:
This addresses #37, #40, and #46.


dgoffredo:
The benchmark job's report generation script is still failing. I reached out to Dima about it. For now I'll skip.

Issue #38, opened 2023-06-02
revise error handling in DatadogAgent


dgoffredo:
Make the handling of response statuses, empty response bodies, etc. consistent with [changes][1] recently made to dd-opentracing-cpp.

[1]: https://github.com/DataDog/dd-opentracing-cpp/pull/260

dgoffredo:
This is addressed.

PR #50, opened 2023-08-24
update unsuccessful HTTP requests handling (#38)

dmehala:
### Content
- consider exclusively 200 as a success HTTP status
- log an error for successfull HTTP status with an empty body

### DIFF
--- a/src/datadog/datadog_agent.cpp
+++ b/src/datadog/datadog_agent.cpp
@@ -229,7 +229,7 @@ void DatadogAgent::flush() {
       logger->log_error([](auto& stream) {
         stream
             << "Datadog Agent returned response without a body."
-               "This tracer might be sending batches of traces too frequently";
+               " This tracer might be sending batches of traces too frequently";
       });
       return;
     }
--- a/src/datadog/datadog_agent.cpp
+++ b/src/datadog/datadog_agent.cpp
@@ -217,8 +217,9 @@ void DatadogAgent::flush() {
                                         std::string response_body) {
     if (response_status != 200) {
       logger->log_error([&](auto& stream) {
-        stream << "Unexpected response status from Datadog Agent"
-               << response_status << " with body (starts on next line):\n"
+        stream << "Unexpected response status " << response_status
+               << " in Datadog Agent response with body of length "
+               << response_body.size() << " (starts on next line):\n"
                << response_body;
       });
       return;
--- a/src/datadog/datadog_agent.cpp
+++ b/src/datadog/datadog_agent.cpp
@@ -215,11 +215,20 @@ void DatadogAgent::flush() {
                       logger = logger_](int response_status,
                                         const DictReader& /*response_headers*/,
                                         std::string response_body) {
-    if (response_status < 200 || response_status >= 300) {
+    if (response_status != 200) {
       logger->log_error([&](auto& stream) {
-        stream << "Unexpected response status " << response_status
-               << " with body (starts on next line):\n"
+        stream << "Unexpected response status from Datadog Agent"
+               << response_status << " with body (starts on next line):\n"
                << response_body;
+      });
+      return;
+    }
+
+    if (response_body.empty()) {
+      logger->log_error([](auto& stream) {
+        stream
+            << "Datadog Agent returned response without a body."
+               "This tracer might be sending batches of traces too frequently";
       });
       return;
     }
--- a/test/test_datadog_agent.cpp
+++ b/test/test_datadog_agent.cpp
@@ -72,8 +72,6 @@ TEST_CASE("CollectorResponse") {
 
     {
       http_client->response_status = 200;
-      http_client->response_body.str("");
-      http_client->response_body.clear();
       Tracer tracer{*finalized};
       auto span = tracer.create_span();
       (void)span;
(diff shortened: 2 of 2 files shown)


src/datadog/datadog_agent.cpp
```diff
@@ -215,15 +215,24 @@ void DatadogAgent::flush() {
                       logger = logger_](int response_status,
                                         const DictReader& /*response_headers*/,
                                         std::string response_body) {
-    if (response_status < 200 || response_status >= 300) {
+    if (response_status != 200) {
       logger->log_error([&](auto& stream) {
-        stream << "Unexpected response status " << response_status
-               << " with body (starts on next line):\n"
+        stream << "Unexpected response status from Datadog Agent"
```
dgoffredo, line 220: Let's put a space after the word "Agent" so that the response status is surrounded by spaces.

src/datadog/datadog_agent.cpp
```diff
@@ -215,15 +215,24 @@ void DatadogAgent::flush() {
                       logger = logger_](int response_status,
                                         const DictReader& /*response_headers*/,
                                         std::string response_body) {
-    if (response_status < 200 || response_status >= 300) {
+    if (response_status != 200) {
```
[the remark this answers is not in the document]
dgoffredo: I agree with you. The special handling of zero was due to the fact that we're using libcurl to do HTTP, and zero can pop up in situations where the actual response status was _not_ zero. However, in this code, I think that those cases would result in an error instead. So, we can ignore any special consideration of zero.

src/datadog/datadog_agent.cpp
dgoffredo: Come to think of it, consider the following order:
```c++
stream << "Unexpected response status " << response_status
       << " in Datadog Agent response with body of length "
       << response_body.size() << " (starts on next line):\n"
       << response_body;
```
Might as well say it all.

test/test_datadog_agent.cpp
```diff
@@ -66,6 +66,23 @@ TEST_CASE("CollectorResponse") {
...
   }
 
+  SECTION("HTTP success with empty body") {
+    // Don't echo error messages.
+    logger->echo = nullptr;
+
+    {
+      http_client->response_status = 200;
+      http_client->response_body.str("");
+      http_client->response_body.clear();
```
dgoffredo, line 76: This testing library we're using, Catch2, has the neat behavior that each `SECTION` is isolated from its cousins.

That is, in the following:
```
TEST_CASE("foo") {
    a;
    SECTION("bar") {
        b;
    }
    c;
    SECTION("baz") {
        d;
        SECTION("bam") {
            e;
        }
        f;
        SECTION("barf") {
            g;
        }
        h;
    }
    i;
}
```
The `TEST_CASE("foo")` will run three times: once for each "leaf" `SECTION` in the tree of nested `SECTION`s.

The order of statements in each run will be:

1. a, b, c, i
2. a, c, d, e, f, h, i
3. a, c, d, f, g, h, i

Pretty neat, right? Gets out of hand pretty quickly, though, with more than a layer or so of nesting.

That's a long way of saying that `http_client->response_body.clear();` is probably not necessary.

test/test_datadog_agent.cpp
```diff
@@ -101,7 +118,7 @@ TEST_CASE("CollectorResponse") {
     // Don't echo error messages.
     logger->echo = nullptr;
 
-    auto status = GENERATE(range(300, 600));
+    auto status = GENERATE(range(201, 600));
```
dgoffredo, line 121: Consider adding a comment here saying why `[201, 299]` are considered errors, even though HTTP says otherwise.

One way to say it is that we know that the Datadog Agent in particular only ever returns 200 on success, and never any of the other 2xx codes. Also there's the corner case where 200 + empty body can mean 429 if the agent is configured in a certain way.

dgoffredo (changes requested):
Looks good! Thanks, Damien. The only changes that I recommend are about the formatting of the error messages.

I also left some other comments for your consideration.

test/test_datadog_agent.cpp
dmehala: That's a really nice and simple introduction de Catch2 I would have loved to have when I encountered it two years ago, it would have saved me so many mistakes. Thank you.

fixed in 8c405b49c68a5c20b5417a1c76261e1e49672b75

src/datadog/datadog_agent.cpp
dmehala: Good catch. Fixed in 8c405b49c68a5c20b5417a1c76261e1e49672b75

test/test_datadog_agent.cpp
dmehala: Added a comment in 8c405b49c68a5c20b5417a1c76261e1e49672b75.

src/datadog/datadog_agent.cpp
```diff
@@ -215,15 +215,25 @@ void DatadogAgent::flush() {
...
                << response_body;
       });
       return;
     }
 
+    if (response_body.empty()) {
+      logger->log_error([](auto& stream) {
+        stream
+            << "Datadog Agent returned response without a body."
+               "This tracer might be sending batches of traces too frequently";
```
dgoffredo, line 232: This is still missing a space between the "body." and the "This".

dgoffredo:
Looking good. Still one error message formatting thing, and then this will be ready to merge.

PR #51, opened 2023-08-24
documentation touchups

dgoffredo:
`doc/maintainers.md` is taking a while, but the rest of these changes are still valuable.


dgoffredo:
I don't think these documentation changes warrant review. Merging...

PR #59, opened 2023-09-28
Feature: Sampling delegation

dmehala:
## How to test

1. `docker compose build --build-arg BRANCH=dmehala/sampling-delegation --no-cache`
2. `[DD_TRACE_DELEGATE_SAMPLING=<0 or 1>] [DD_API_KEY=<API_KEY>] docker compose up`
3. Enjoy! 🥳

PR #60, opened 2023-10-12
sampling delegation: code review ideas from #59

dgoffredo:
This is more than I set out to write, but sampling has a way of dragging me into the pit.

My code is uglier than I anticipated. Maybe some of this complexity can be reduced, or maybe it's intrinsic.

Some of these changes are "just, like, my opinion, man," and others are real changes to how we keep track of the "delegation state." See the commits for a slightly-finer-grained breakdown of what I propose.

The big `TEST_CASE` that I added is the crux of it. There are yet still other corner cases to consider.

Let's talk through this in our next meeting.


src/datadog/trace_segment.cpp
```diff
@@ -337,9 +362,9 @@ bool TraceSegment::inject(DictWriter& writer, const SpanData& span,
         if (origin_) {
           writer.set("x-datadog-origin", *origin_);
         }
-        if (opts.delegate_sampling_decision) {
-          expecting_response = true;
-          expecting_delegated_sampling_decision_ = true;
+        if (options.delegate_sampling_decision) {
+          delegated_trace_sampling_decision =
+              sampling_delegation_.sent_request_header = true;
```
dgoffredo, line 367: Oops. It would be better if I didn't use assignment chaining:
```c++
    delegated_trace_sampling_decision = true;
    sampling_delegation_.sent_request_header = true;
```

src/datadog/trace_segment.cpp
```diff
@@ -389,37 +414,32 @@ void TraceSegment::write_sampling_delegation_response(DictWriter& writer) {
...
   }
 
-  if (expecting_delegated_sampling_decision_) {
-    expecting_delegated_sampling_decision_ = false;
-    trace_tags_.emplace_back(tags::internal::sampling_decider, "1");
-  }
-
-  writer.set(DD_SAMPLING_DELEGATION_RESPONSE_HEADER, j.dump());
-  return;
+  writer.set(sampling_delegation_response_header, j.dump());
```
dgoffredo, line 417: meeting with Damien: Maybe we should decide on a "static" constant naming convention. `DD_CAPS_STUFF` might work, but looks like an env variable. `some_thing` looks like a local. Maybe `k_foo_bar` or similar?

src/datadog/trace_segment.cpp
```diff
@@ -389,37 +414,32 @@ void TraceSegment::write_sampling_delegation_response(DictWriter& writer) {
...
-      }
-    }
+  sampling_delegation_.received_matching_response_header = true;
+
+  std::lock_guard<std::mutex> lock(mutex_);
+  // Overwrite any existing sampling decision if and only if the existing
+  // decision is not a local manual override.
+  if (!(sampling_decision_ &&
+        sampling_decision_->origin == SamplingDecision::Origin::LOCAL &&
+        sampling_decision_->mechanism == int(SamplingMechanism::MANUAL))) {
```
dgoffredo, line 440: Feel free to put the `static_cast` back (instead of my function-style C cast).

test/test_tracer.cpp
```diff
@@ -1210,3 +1220,207 @@ TEST_CASE(
...
+
+  auto valid_config = finalize_config(config1);
+  REQUIRE(valid_config);
+  Tracer tracer1{*valid_config};
+
+  valid_config = finalize_config(config2);
+  REQUIRE(valid_config);
+  Tracer tracer2{*valid_config};
+
+  valid_config = finalize_config(config3);
```
dgoffredo, line 1284: need a `REQUIRE(valid_config)` here

test/test_tracer.cpp
```diff
@@ -1210,3 +1220,207 @@ TEST_CASE(
...
+    REQUIRE(delegation_writer.items.empty());
+  }
+
+  // three segments, each having two spans
+  REQUIRE(collector->span_count() == 3 * 2);
+
+  const double expected_sampling_priority = double(SamplingPriority::USER_KEEP);
+  // "dm" as in the "_dd.p.dm" tag
+  const std::string expected_dm =
+      "-" + std::to_string(int(SamplingMechanism::RULE));
```
dgoffredo, line 1365: `static_cast`?

src/datadog/trace_segment.cpp
dgoffredo: Nah, not for numeric and enum conversions. We can revisit this across the entire repo another time. For now I'll keep in consistent with other uses.

Compare the output of the following two commands:
```
$ git grep '\<int('
``` 
```
$ git grep static_cast src/ ':!*json.hpp'
```

src/datadog/trace_segment.cpp
dgoffredo: We can decide later.

PR #62, opened 2023-10-12
Implement telemetry API v2

cgilmour:
This implements an internal API that measures activity within the tracer and reports it to the Datadog Agent.
Users are able to disable this via the environment variable `DD_INSTRUMENTATION_TELEMETRY_ENABLED` or the `report_telemetry` config option in `TracerConfig`.

Additional metrics may be implemented in the future.

### DIFF
--- a/src/datadog/datadog_agent.cpp
+++ b/src/datadog/datadog_agent.cpp
@@ -339,8 +339,8 @@ void DatadogAgent::send_app_started(nlohmann::json&& tracer_config) {
       telemetry_endpoint_, telemetry_set_request_headers_, std::move(payload),
       telemetry_on_response_, telemetry_on_error_);
   if (auto* error = post_result.if_error()) {
-    logger_->log_error(
-        error->with_prefix("Unexpected error submitting telemetry app-started event: "));
+    logger_->log_error(error->with_prefix(
+        "Unexpected error submitting telemetry app-started event: "));
   }
 }
 
@@ -350,8 +350,8 @@ void DatadogAgent::send_heartbeat_and_telemetry() {
       telemetry_endpoint_, telemetry_set_request_headers_, std::move(payload),
       telemetry_on_response_, telemetry_on_error_);
   if (auto* error = post_result.if_error()) {
-    logger_->log_error(
-        error->with_prefix("Unexpected error submitting telemetry app-heartbeat event: "));
+    logger_->log_error(error->with_prefix(
+        "Unexpected error submitting telemetry app-heartbeat event: "));
   }
 }
 
@@ -361,8 +361,8 @@ void DatadogAgent::send_app_closing() {
       telemetry_endpoint_, telemetry_set_request_headers_, std::move(payload),
       telemetry_on_response_, telemetry_on_error_);
   if (auto* error = post_result.if_error()) {
-    logger_->log_error(
-        error->with_prefix("Unexpected error submitting telemetry app-closing event: "));
+    logger_->log_error(error->with_prefix(
+        "Unexpected error submitting telemetry app-closing event: "));
   }
 }
 


src/datadog/datadog_agent.cpp
```diff
@@ -211,10 +265,21 @@ void DatadogAgent::flush() {
 
   // This is the callback for the HTTP response.  It's invoked
   // asynchronously.
-  auto on_response = [samplers = std::move(response_handlers),
+  auto on_response = [this, samplers = std::move(response_handlers),
```
dgoffredo, line 268: I'm thinking to err on the side of `[telemetry = tracer_telemetry_, ...]` instead of `[this, ...]`.

On the other hand, we can go all in on "the DatadogAgent will outlive the HTTPClient" and use `this` everywhere, without copying any of `*this`'s members. The lifetimes are correct.

The reason I prefer the unnecessary `shared_ptr` copies is that it might aid in a future refactoring of `class Curl` into different "event loop" backends.

Anyway, let's go with one or the other. Which is up to you.

src/datadog/datadog_agent.cpp
```diff
@@ -263,5 +330,35 @@ void DatadogAgent::flush() {
   }
 }
 
+void DatadogAgent::send_app_started(nlohmann::json&& tracer_config) {
+  auto payload = tracer_telemetry_->app_started(std::move(tracer_config));
+  auto post_result = http_client_->post(
+      telemetry_endpoint_, telemetry_set_request_headers_, std::move(payload),
+      telemetry_on_response_, telemetry_on_error_);
+  if (auto* error = post_result.if_error()) {
+    logger_->log_error(*error);
```
dgoffredo, line 339: Reconsider using `error.with_prefix` here and below, as you did above (to distinguish telemetry stuff from trace stuff, and to distinguish among the telemetry stuff).

src/datadog/metrics.h
```diff
@@ -0,0 +1,71 @@
...
+#include <vector>
+
+namespace datadog {
+namespace tracing {
+
+class Metric {
+  // The name of the metric that will be published. A transformation occurs
+  // based on the name and whether it is "common" or "language-specific" when it
+  // is recorded.
+  const std::string name_;
```
dgoffredo, line 20: `const`? _const_? Is this data member... immutable? _Is that why you used `const`?_

Still not local, so fair game.

dgoffredo (commented):
Excellent!

Here's my first pass through the changes. See my comments inline. Overall:

- good documentation
- design and style are consistent with the rest of the library
- test coverage is almost perfect on a line-by-line basis. You [missed a spot][1].

I'll do another pass where I look at the unit tests in more detail, and the serialization code. And maybe poke around with an Agent proxy.

This is pretty much ready to merge, once we come to an agreement on ideas raised in the comments.

[1]: https://datadoghq.dev/dd-trace-cpp-coverage/2023-10-12%2019:41:48%20UTC%20(04e1a17)/datadog/metrics.cpp.gcov.html

src/datadog/datadog_agent.cpp
```diff
@@ -346,7 +350,8 @@ void DatadogAgent::send_heartbeat_and_telemetry() {
       telemetry_endpoint_, telemetry_set_request_headers_, std::move(payload),
       telemetry_on_response_, telemetry_on_error_);
   if (auto* error = post_result.if_error()) {
-    logger_->log_error(*error);
+    logger_->log_error(
+        error->with_prefix("Unexpected error submitting traces: "));
```
dgoffredo, line 354:
```suggestion
        error->with_prefix("Unexpected error submitting telemetry data: "));
```

src/datadog/datadog_agent.cpp
```diff
@@ -336,7 +339,8 @@ void DatadogAgent::send_app_started(nlohmann::json&& tracer_config) {
       telemetry_endpoint_, telemetry_set_request_headers_, std::move(payload),
       telemetry_on_response_, telemetry_on_error_);
   if (auto* error = post_result.if_error()) {
-    logger_->log_error(*error);
+    logger_->log_error(
+        error->with_prefix("Unexpected error submitting telemetry: "));
```
dgoffredo, line 343:
```suggestion
        error->with_prefix("Unexpected error submitting telemetry app-started event: "));
```

src/datadog/metrics.cpp
```diff
@@ -5,24 +5,24 @@
 namespace datadog {
 namespace tracing {
 
-Metric::Metric(const std::string name, std::string type,
-               const std::vector<std::string> tags, bool common)
+Metric::Metric(std::string name, std::string type,
+               std::vector<std::string> tags, bool common)
     : name_(name), type_(type), tags_(tags), common_(common) {}
```
dgoffredo, line 10: Since we're accepting these parameters by value, might as well `std::move` them into these initializers.

Alternatively, use `const T&` instead.

Alternatively, use `const vector&` and `StringView`.

Alternatively, use `std::move(vector)` and `StringView`.

It's true that this code runs on tracer startup only, so it doesn't matter. But, style.

dgoffredo: You might also just leave it like this. Less ink.

dgoffredo: Now that I'm looking at the individual commits again, I wonder if the original `const` that you had on these parameters was a typo where you were missing the `&`. Anyway, figure out what you want the signatures and initializers to look like, based on how the constructor is called.

dgoffredo (approved):
Looks good to me.

Thinking a bit more about the (remote?) possibility that this code gets deployed to a system with a Datadog Agent that's old enough not to have telemetry, do consider silencing or perhaps one-timing the error in `telemetry_on_response_` when the response status is 404 ("error: Datadog Agent does not look like it supports telemetry", then subsequent silence), or something like that. Also reasonable would be to leave it as is.

Please see my comments about the type signatures of `Metric`'s constructor parameters.

Aside from those ideas, ship it.

src/datadog/datadog_agent.cpp
```diff
@@ -22,13 +22,20 @@ namespace tracing {
...
 
 HTTPClient::URL traces_endpoint(const HTTPClient::URL& agent_url) {
   auto traces_url = agent_url;
   append(traces_url.path, traces_api_path);
   return traces_url;
 }
 
+HTTPClient::URL telemetry_endpoint(const HTTPClient::URL& agent_url) {
+  auto telemetry_v2_url = agent_url;
+  append(telemetry_v2_url.path, telemetry_v2_path);
```
dmehala, line 35:
```suggestion
HTTPClient::URL telemetry_endpoint(HTTPClient::URL telemetry_v2_url) {
  append(telemetry_v2_url.path, telemetry_v2_path);
```

dmehala: If you accept, it should be also done for `traces_endpoint` for consistency

src/datadog/datadog_agent.cpp
```diff
@@ -124,24 +131,71 @@ std::variant<CollectorResponse, std::string> parse_agent_traces_response(
...
+      telemetry_endpoint_(telemetry_endpoint(config.url)),
       http_client_(config.http_client),
       event_scheduler_(config.event_scheduler),
       cancel_scheduled_flush_(event_scheduler_->schedule_recurring_event(
           config.flush_interval, [this]() { flush(); })),
       flush_interval_(config.flush_interval) {
   assert(logger_);
+  assert(tracer_telemetry_);
+  if (tracer_telemetry_->enabled()) {
+    // Only schedule this if telemetry is enabled.
```
dmehala, line 151: Details but the line above is self explanatory.

src/datadog/metrics.h
```diff
@@ -0,0 +1,69 @@
...
+         bool common);
+
+ public:
+  // Accessors for name, type, tags, common and capture_and_reset_value are used
+  // when producing the JSON message for reporting metrics.
+  std::string name();
+  std::string type();
+  std::vector<std::string> tags();
+  bool common();
+  uint64_t value();
```
dmehala, line 42: I will probably be controversial here but we could get rid of those accessors 👀

dgoffredo: `value()` hides the atomicity of the underlying member, for better or for worse.

My vote is to keep the trivial methods, but I thought the same as you when I initially read it.

src/datadog/runtime_id.h
```diff
@@ -0,0 +1,28 @@
...
+// which this tracing library is embedded.
+//
+// See `TracerConfig::runtime_id`, declared in `tracer_config.h`.
+
+#include <string>
+
+namespace datadog {
+namespace tracing {
+
+class RuntimeID {
```
dmehala, line 14: From my understanding, `RuntimeID` is a `std::string` in a tuxedo. How about `using RuntimeID = std::string` and `generate()` as free function? We lose strong typing in the process but we don't use it for now.

src/datadog/tracer_telemetry.cpp
```diff
@@ -0,0 +1,211 @@
...
+    bool enabled, const Clock& clock, const std::shared_ptr<Logger>& logger,
+    const std::shared_ptr<const SpanDefaults>& span_defaults,
+    const RuntimeID& runtime_id)
+    : enabled_(enabled),
+      clock_(clock),
+      logger_(logger),
+      span_defaults_(span_defaults),
+      runtime_id_(runtime_id),
+      hostname_(get_hostname().value_or("hostname-unavailable")) {
+  if (enabled_) {
```
dmehala, line 21: It is probably better to move that logic in the caller, so depending of `enabled` it can instantiate a `NoopTracerTelemetry` or this implementation.

src/datadog/runtime_id.h
dgoffredo: That's similar to what Caleb had before.

The reason I dressed our string in formal wear is because we were exposing the concept of RuntimeID to the user for the first time. This strict technique guarantees that RuntimeID is a UUID. Not that anybody was likely to set it to anything else...

src/datadog/tracer_telemetry.cpp
```diff
@@ -0,0 +1,211 @@
...
+        // gauge metrics have a interval
+        metrics.emplace_back(nlohmann::json::object({
+            {"metric", metric.name()},
+            {"tags", metric.tags()},
+            {"type", metric.type()},
+            {"interval", 10},
+            {"points", points},
+            {"common", metric.common()},
+        }));
+      }
```
dmehala, line 146: That could be done while capturing metrics so we avoid iterating twice every 60s and it will make `heartbeat_and_telemetry` easier.

src/datadog/tracer_telemetry.cpp
dgoffredo: I'm thinking about what a `TracerTelemetry` _interface_ would look like. What would `metrics()` return?

I prefer to leave this concrete as Caleb has it, but Damien's suggestion is worth thinking about more.

src/datadog/tracer_telemetry.cpp
dgoffredo: I think we don't even use gauge currently. I assume Caleb left it in anticipating it would be used soon. Technically dead code, though, what do you think?

dmehala (approved):
LGTM :shipit:

PR #63, opened 2023-10-16
prettier test coverage reporting

dgoffredo:
This includes the git branch name, along with a summary of the coverage percentages, in the name of each subdirectory.


dgoffredo:
I merged this for now so that subsequent CI jobs don't break the coverage report (not that it'd be too big a deal).

This is still fair game for review.

PR #64, opened 2023-10-18
Report tracer_version instead of tracer_version_string

cgilmour:
Report `v1.2.3` instead of `[dd-trace-cpp version v1.2.3]`.
The latter value doesn't work nicely in telemetry dashboards.


dmehala (approved):
LGTM :shipit: Would it make sense to add a unit test?

dgoffredo (approved):
This was a typo on my part. I don't think it needs a test.

PR #65, opened 2023-10-19
curl minutiae

dgoffredo:
There's no need to hold the lock while executing the response handler. It's not an issue today, but it does prevent a handler from submitting another request.


dmehala (approved):
:shipit:

PR #66, opened 2023-10-19
Set default timeout on curl handles

dmehala:
### Why?
Now, `drain` guarantee after being executed all ongoing requests are processed (unless requests are posted while draining 👀) and if for whatever reason the request could not be processed, like timeout, the reason will be logged.

Content:
- Curl default implementation set a 2s default timeout.
- Drain waits indefinitely until the predicate is valid.


dgoffredo:
### `class HTTPClient` Use Cases

1. Envoy

   Doesn't use libcurl. Request timeout of 1 second is hard-coded in implementation.
`drain()` does nothing because the implementation is not threaded.

2. nginx-datadog

   The motivating use case. The purpose of `drain()` is to guarantee that dd-trace-cpp I/O does not block worker shutdown for too long.
   It uses the `drain()` timeout hard-coded into class DatadogAgent.

3. `class DatadogAgent`

   Flushes batches of trace segments every two seconds (by default). It's error handler logs, and its success handler interprets the response for sampling configuration.
   drain() is called in the destructor, so that when the tracer/application is shutting down, the last two seconds of trace segments have a chance to be sent to the agent.
   DatadogAgent doesn't specify a timeout in its requests, because timeouts are not part of the HTTPClient interface.

4. Custom code using `DatadogAgent`

   Aside from Envoy and nginx-datadog, a hypothetical customer might instrument their own C++ code using dd-trace-cpp. They would be instantiating `class Tracer`, which implicitly instantiates `class DatadogAgent`, which implicitly creates a `class Curl`.
   Currently, the only thing that such a user can control is the flush interval. They can't adjust libcurl's request timeouts, and they can't adjust the timeout for `drain()`.

5. Custom code using `HTTPClient`

   A even more hypothetical customer might use `class Curl` for other purposes, in addition to sending traces to the agent. Such a customer can send requests to any endpoint, but cannot specify a timeout. They can `drain()` with a deadline, though.

### Use cases in light of these changes

It doesn't make any difference to Envoy.

The only difference to nginx-datadog is that timeouts at any time will log an error, instead of quietly failing only when `drain()` is called. This is the reason for your changes.

The difference for `class DatadogAgent` is that it can no longer guarantee an upper bound on how long its destructor might block, though with knowledge of the particular implementation of the underlying `HTTPClient`, it will typically be two seconds.

The difference for direct use of `class Curl` is that there is now a hard-coded timeout, and `drain()` will block for no longer than that timeout.

### My Recommendation

Let's add all of the bells and whistles.

Here's an idea:

- Add a `std::chrono::steady_clock::time_point deadline` parameter to `HTTPClient::post`.
- Convey that `deadline` in `struct CurlImpl::Request`. Don't convert it into a `CURLOPT_TIMEOUT_MS` until the moment before the handle is added to the multi handle by the `run()` thread.
- Add an `int request_timeout_milliseconds = 2000` member to `struct DatadogAgentConfig`, and a corresponding `std::chrono::steady_clock::duration request_timeout` to `FinalizedDatadogAgentConfig`.
- Add an `int shutdown_timeout_milliseconds = 2000` member to `struct DatadogAgentConfig`, and a corresponding `std::chrono::steady_clock::duration shutdown_timeout` to `FinalizedDatadogAgentConfig`.
- Restore the deadline for `drain()`.
- Have `~DatadogAgent()` use its configured shutdown timeout to calculate a deadline for `drain()`.
- Have `DatadogAgent::flush()` use its configured request timeout to calculate a deadline for `HTTPClient::post`.

That's the maximalist solution. What you have proposed in this PR is an improvement over the status quo in the sense that agent request taking more than two seconds log an error, which is good, but the omission of a deadline from the signature of `drain()` degrades the appearance of multi-purpose interfaces. I haven't thought of a middle ground between what you propose and the all-out modifications described above.

What do you think?

src/datadog/curl.cpp
```diff
@@ -464,6 +473,7 @@ CURLMcode CurlImpl::log_on_error(CURLMcode result) {
 void CurlImpl::run() {
   int num_messages_remaining;
   CURLMsg *message;
+  const int max_wait_milliseconds = 1000;
```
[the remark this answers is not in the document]
dmehala: That's from another experimentation. Back when my hypothesis was `drain()` has been called but the thread is waiting for an event from curl or that previous 10s timeout. In that scenario, `drain()` would be useless. Is it happening ? I don't know 🤷‍♂️ One thing I am sure of is it is not here on purpose, so good and catch and:
- [ ] revert

dmehala:
> Custom code using `DatadogAgent`
They would be instantiating `class Tracer`, which implicitly instantiates `class DatadogAgent`, which implicitly creates a `class Curl`

The consensus is that library is only used by us for webserver integrations. I do not know if we want to be more than that, but if we do, it will useful to let users of the library make use of their own transport library.

Your recommandation is going deeper and makes everything more explicit. Overall, I do like it even if I am still not convinced the removal of a deadline in `drain()` degrades the appearance of multi-purpose interfaces. 

All in all, my only suggestion is to increase the shutdown timeout. It must always be bigger than the request timeout to actually give a chance for the request to be processed. Otherwise, in some cases (shutdown) it will makes no difference with our current solution. I may be wrong, what's your thought?

src/datadog/curl.cpp
```diff
@@ -368,7 +369,7 @@ Expected<void> CurlImpl::post(const HTTPClient::URL &url,
   throw_on_error(curl_.easy_setopt_headerdata(handle.get(), request.get()));
   throw_on_error(curl_.easy_setopt_writefunction(handle.get(), &on_read_body));
   throw_on_error(curl_.easy_setopt_writedata(handle.get(), request.get()));
-  throw_on_error(curl_.easy_setopt_timeout(handle.get(), k_default_timeout_ms));
+  throw_on_error(curl_.easy_setopt_timeout(handle.get(), timeout.count()));
```
dmehala, line 372: Use `std::chrono::duration_cast<std::chrono::milliseconds>()` instead because the timeout can be in another unit

src/datadog/datadog_agent_config.cpp
```diff
@@ -114,6 +114,22 @@ Expected<FinalizedDatadogAgentConfig> finalize_config(
   result.flush_interval =
       std::chrono::milliseconds(config.flush_interval_milliseconds);
 
+  if (config.request_timeout_milliseconds <= 0) {
+    return Error{Error::DATADOG_AGENT_INVALID_FLUSH_INTERVAL,
```
dmehala, line 118: The error code is not the right one because I would like to get your thought. How about a less specific error code? Like `DATADOG_AGENT_INVALID_CONFIGURATION`?

src/datadog/http_client.h
```diff
@@ -46,11 +47,11 @@ class HTTPClient {
   // preparing the request, return an `Error`.
   virtual Expected<void> post(const URL& url, HeadersSetter set_headers,
                               std::string body, ResponseHandler on_response,
-                              ErrorHandler on_error) = 0;
+                              ErrorHandler on_error, Timeout timeout) = 0;
 
   // Wait until there are no more outstanding requests, or until the specified
   // `deadline`.
-  virtual void drain() = 0;
+  virtual void drain(std::chrono::steady_clock::time_point deadline) = 0;
```
dmehala, line 54: I would prefer `drain(Timeout timeout)` instead but that requires passing the clock in the ctor and changing too many part of the codebase. Good side effect would be `test_curl` would use a fake clock.

test/test_curl.cpp
```diff
@@ -15,6 +15,7 @@
 #include "test.h"
 
 using namespace datadog::tracing;
+using namespace std::chrono_literals;
```
dmehala, line 18: - [ ] revert

test/test_curl.cpp
```diff
@@ -182,7 +189,8 @@ TEST_CASE("parse response headers and body") {
             exception = std::current_exception();
           }
         },
-        [&](const Error &error) { post_error = error; });
+        [&](const Error &error) { post_error = error; },
+        std::chrono::seconds(0));
```
dmehala, line 193: - [ ] set to `std::chrono::seconds(10)`

src/datadog/datadog_agent_config.cpp
dgoffredo: I like having specific error messages. In most cases, when I'm writing new code and encounter an error condition, I add a new value to `enum class Error::Code`.

test/test_curl.cpp
```diff
@@ -251,12 +257,14 @@ TEST_CASE("fail to allocate request handle") {
 
   const auto ignore = [](auto &&...) {};
   const HTTPClient::URL url = {"http", "whatever", ""};
-  const auto result = client->post(url, ignore, "dummy body", ignore, ignore);
+  const auto ignore_timeout = std::chrono::seconds(10);
```
dgoffredo, line 260: maybe `arbitrary_timeout` or `dummy_timeout`.

dgoffredo (approved):
Since my suggestions were merged, I approve of these changes.

Hopefully the merge conflicts aren't too much trouble to sort out.

dgoffredo (approved):

PR #73, opened 2023-12-01
Use CMake FetchContent for `curl`

dmehala:
## Description

Rework `curl` dependency. Now, we use `FetchContent` which populates dependencies during the configuration stage and bringing all the targets declared by `curl`.

## How does it fix DataDog/nginx-datadog#53?

The build system is now able to bring transitive dependencies into all targets that declare `curl` as a dependency.

## Note to reviewer
I also added the equivalent of `--without-zstd`.


CMakeLists.txt
dmehala, line 1: - [ ] Remove `MAKE_JOB_COUNT` variable

dgoffredo (approved):
Nice!

The resulting `.so`, on my system, is larger now.

Before:

    -rwxrwxr-x 1 david david 21943168 Dec  1 11:20 .build/libdd_trace_cpp.so

After:

    -rwxrwxr-x 1 david david 23314640 Dec  1 11:18 .build/libdd_trace_cpp.so

I wonder if this is due to the CMake-flavored libcurl build including more code than the configure-based one.

As we've discussed before, the only concern I have with these changes is that they allow dependencies to creep into the build, unnecessary and unnoticed.

I think it's fine, though. This is good to merge.

PR #74, opened 2023-12-04
Add support for APM Remote Configuration

dmehala:
## Description

Remote Configuration allows to remotely configure the library. For now, it is only possible to update the trace sampling rate remotely.

## How to test

An API Key with Remote Configuration enabled is required.

1. Run the `hasher` example.
2. Generate some traces so the service can appear in the Service Catalogue.
3. In Datadog's UI go to the Service Catalogue and select the `dd-trace-cpp-example` service.
4. An APM Instrumentation banner should be visible.

output_trimmed_enhanced


dgoffredo:
I'm playing with my own branch, making suggestions for this PR.

In looking at `TracerId` and the related parts of the remote config API, I noticed the following paragraph from our internal "[RFC] Integrating with Remote Config in a Tracer":

> The Client.ClientState field in  `ClientGetConfigsRequest` MUST contain the global state of RC in the tracer. There MUST NOT exist separate instances of RC clients in a tracer making separate `ClientGetConfigsRequest` with their own separated `Client.ClientState`. We use this state to track individual tracers in the agent and the backend. Having multiple components in the same tracer each call this API with separated state will duplicate a lot of data and cause issues in the attribution of configurations.

What is meant by "separate" here? My guess is that each `Client.ClientState` is specific to _a unique tracer_. What then determines the identify of a "tracer"? Is it just `client.id`? Is it `client.id` combined with `client.client_tracer.*`, as you have largely done in defining `struct TracerId`?

The notion of "tracer ID" is suspect in my mind. How does it all fit together?

dmehala:
> What is meant by "separate" here? My guess is that each Client.ClientState is specific to a unique tracer. What then determines the identify of a "tracer"? Is it just client.id? Is it client.id combined with client.client_tracer.*, as you have largely done in defining struct TracerId?

I asked RC people, here are there answer:
> `client.id` is used to identify individual tracers
There is also assumptions being made such as "it's impossible to have two different client.id for a single client.runtimeID". Breaking those assumptions will likely break some UIs and potentially confuse systems that distribute configurations to specific tracers

From my understanding it is a combination `client.id` and `client.client_tracer.runtime_id`. If I remember they should not be the same but one `client.id` is associated to one `client.client_tracer.runtime_id`. Did I missed something?

> The notion of "tracer ID" is suspect in my mind. How does it all fit together?
My main objective in incorporating the TracerID data structure was to enhance code simplicity and improve expressiveness.

dgoffredo:
> There is also assumptions being made such as "it's impossible to have two different client.id for a single client.runtimeID".

Currently, each `class Tracer` creates its own `RuntimeID`. When envoy or nginx creates worker threads/processes, each gets its own `Tracer`, and so each gets its own `RuntimeID`.

@cgilmour  and I decided that it would be better if these workers shared the same `RuntimeID`. They are, after all, part of the same runtime. The first bit of code that would enforce this are the [proposed fixes][1] to our envoy code that enable newer versions of dd-trace-cpp. See `DatadogTracerFactory::makeConfig` in that change set.

I'm concerned that now each worker will have _the same_ "tracer ID," (same runtime, env, service) but that the workers' `client.id` will all be different. `client.id` is generated once per `Tracer`.

What I'm trying to get at, asking these questions, is _what is a tracer?_ Worst case scenario, there is no clear definition.

Here's my opinion: Workers within a proxy are the "same tracer." They're part of the same runtime and _share the same configuration_. So, I figure each of these workers should appear identical to the remote config backend. Will that be a problem? I think it will be a problem.

Alternatively, workers can share a runtime ID, env, and service, but have _different client IDs_. What is meant by "client"? What is the relationship between "client" and "tracer"?

Let's iron out these definitions before committing the code. At the very least, it will allow us to write good documentation for the `TracerID` component.

[1]: https://github.com/DataDog/envoy/pull/6

src/datadog/datadog_agent_config.cpp
```diff
@@ -117,21 +118,28 @@ Expected<FinalizedDatadogAgentConfig> finalize_config(
...
-    return Error{Error::DATADOG_AGENT_INVALID_SHUTDOWN_TIMEOUT,
-                 "DatadogAgent: Shutdown timeout must be a positive number of "
-                 "milliseconds."};
+  if (rc_poll_interval_seconds <= 0) {
+    return Error{Error::DATADOG_AGENT_INVALID_REMOTE_CONFIG_POLL_INTERVAL,
+                 "DatadogAgent: Remote Configuration poll interval must be a "
+                 "positive number of seconds."};
   }
-  result.shutdown_timeout =
-      std::chrono::milliseconds(config.shutdown_timeout_milliseconds);
```
dmehala, line 134: - [ ] fix rebase

dgoffredo (commented):
Looking good. Please see my suggestions in <https://github.com/DataDog/dd-trace-cpp/pull/83>.

Once we've discussed that, I'll mark this approved.

dmehala:
@dgoffredo I added your changes in 8a9d59b621b42ed54e9d0982b3caf5837f6e2472 and 344424506dcaa64dc971625b82dec77e860f54e5

src/datadog/tracer_signature.h
```diff
@@ -0,0 +1,53 @@
...
+//    `RemoteConfigurationManager::process_response` in `remote_config.h`.
+
+#include <string>
+
+#include "runtime_id.h"
+#include "string_view.h"
+#include "version.h"
+
+#define DD_TRACE_STRINGIFY(ARG) DD_TRACE_STRINGIFY_HELPER(ARG)
+#define DD_TRACE_STRINGIFY_HELPER(ARG) #ARG
```
dgoffredo, line 29: You can `#undef` these after use, if you like.

dgoffredo (approved):
Very nice! This feature came out well.

Once this is merged, I'll see whether it breaks Envoy's unit tests. We can keep the two concerns decoupled.

PR #76, opened 2023-12-07
review of dmehala/remote-config-impl (#74)

dgoffredo:
Each commit in this pull request is effectively a review comment for <https://github.com/DataDog/dd-trace-cpp/pull/74>.

This pull request is a work in progress.

So far, my suggestions are all stylistic.

Please note specifically any `TODO(@dgoffredo)` comments that I've added.


src/datadog/remote_config.cpp
```diff
@@ -138,18 +149,16 @@ void RemoteConfigurationManager::process_response(const nlohmann::json& json) {
 
       const auto& config_metadata =
           targets.at("/signed/targets"_json_pointer).at(config_path);
-      if (!contains(config_path, k_apm_product) ||
+      if (!contains(config_path, k_apm_product_path_substring) ||
```
dgoffredo, line 152: Idea from Caleb: Write a partial parser that reads the first path component, and based on its value either retrieves the appropriate component, or returns some "n/a" value.

src/datadog/base64.cpp
```diff
@@ -1,109 +1,100 @@
...
+      c3 == k_sentinel) {
     return "";
   }
 
-  decoder.buffer = static_cast<uint32_t>(0) | static_cast<uint32_t>(c0) << 26 |
-                   static_cast<uint32_t>(c1) << 20 |
-                   static_cast<uint32_t>(c2) << 14 |
-                   static_cast<uint32_t>(c3) << 8;
+  decoder.buffer = 0 | c0 << (32 - 6 * 1) | c1 << (32 - 6 * 2) |
+                   c2 << (32 - 6 * 3) | c3 << (32 - 6 * 4);
```
dmehala, line 83: I don't find all the arithmetic particularly helpful, it does the opposite of what intended IMO. Could not make that change?

dmehala (approved):
Except the two comment this looks good to me. Thank you for fixing those little details I forgot like the documentation, better comment, moving stuff to `.cpp` and more.

src/datadog/config_manager.h
```diff
@@ -10,54 +12,32 @@
...
 
  public:
-  ConfigManager(const FinalizedTracerConfig& config)
-      : clock_(config.clock),
-        default_trace_sampler_(
-            std::make_shared<TraceSampler>(config.trace_sampler, clock_)),
-        current_trace_sampler_(default_trace_sampler_) {}
+  struct Update {
+    Optional<TraceSamplerConfig> trace_sampler;
+  };
```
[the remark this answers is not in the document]
dgoffredo: I saw three options:

1. Leave it as you had it.
2. Put `struct ConfigUpdate` in its own component (its own `.h`). Good place for documentation!
3. Consider `ConfigUpdate` as part of `ConfigManager`'s public interface.

(1) and (3) are compatible, but so too are (3) and my proposed change.

My reasoning was "if I see `ConfigUpdate` used in `RemoteConfigManager`, where do I go looking for its definition?" I might look for a `config_update.h`, but there is none. `ConfigUpdate` is a parameter to `ConfigManager`, so I might look there, but the connection is even stronger if it's spelled `ConfigManager::Update`.

> It can also be interpreted as an `Update` of the `ConfigManager`, which is not the intended meaning.

That's a good point. I prefer (2) to (1) if you're looking for a compromise. I'm also fine with (1). This PR is illustrative only.

dgoffredo (commented):
Thanks for taking a look. I left some responses as inline comments.

Thoughts:

- If you disagree with my style suggestions, go with your gut. This is your rodeo.
- Please see my comment mentioning "Idea from Caleb." This is related to the griping I did on slack yesterday about the config path concept.
- One opinion that I _do_ hold strongly is endian agnosticism. Please choose an alternative base64 decoding algorithm. I won't be picky about the style.

dmehala:
Squash and cherry-picked 52375e0c48269eb755f4dec5e90a94bc71fe0f9a in #74

PR #77, opened 2023-12-11
Part 1 - Active configuration

dmehala:
Preparatory work for Active Configuration.

- User configuration uses `std::optional`. It allows use to know if value is the default value or from the user. That distinction is necessary for reporting the configuration for telemetry purposes.


dmehala:
Got feedbacks which lead to another idea.

PR #78, opened 2023-12-18
[Part 1] Active configuration

dmehala:
Define lean interface


src/datadog/tracer_config.h
```diff
@@ -27,156 +30,217 @@ class Logger;
...
+  //
+  // Overriden by the `DD_TRACE_AGENT_URL` environment variable.
+  TracerConfig& set_datadog_agent_url(std::string url);
+
+  // Indicates whether traces generated by the tracer will be
+  // sent to a collector (`true`) or discarded on completion (`false`).
+  // If `report_traces` is `false`, then both `agent` and `collector` are ignored.
+  //
+  // Overridden by the `DD_TRACE_ENABLED` environment variable.
+  TracerConfig& enable_traces(bool report_traces);
```
dgoffredo, line 119: `report_traces` might be a better name, as we discussed.

dgoffredo (approved):
Per our most recent meeting:

- Let's not have `FinalizeTracerConfig` as a data member within `TracerConfig`.
- Let's change the name of the 128-bit trace ID config option.

Other than that, this looks good. Looking forward to how you decide to track config origin, etc.

PR #81, opened 2023-12-18
[ci-chore] Add system tests to our pipeline

dmehala:


.circleci/config.yml
```diff
@@ -90,3 +136,8 @@ workflows:
             arch: ["amd64", "arm64"]
             toolchain: ["gnu", "llvm"]
     - coverage
+    - system-tests:
+        filters:
+          branches:
+            only:
+              - main
```
dmehala, line 143: We might want to revisit this filter

.circleci/config.yml
```diff
@@ -73,6 +73,52 @@ jobs:
...
+            pyenv local system
+            python3.9 --version
+            python3.9 -m pip install wheel
+            python3.9 -m pip install -r requirements.txt
+            sudo ln -sf /usr/bin/python3.9 /usr/bin/python
+      - run:
+          name: Run
+          environment:
+            TEST_LIBRARY: cpp
+            DD_TRACE_CPP_COMMIT: << pipeline.git.revision >>
```
dmehala, line 112: - [ ] Update `system-tests` to build the cpp server on a specific commit id.

.circleci/config.yml
```diff
@@ -73,6 +73,52 @@ jobs:
...
     - run: bin/publish-coverage
 
+  # Copy-pasta from dd-trace-java. Thank you <3
+  system-tests:
+    parameters:
+      systemTestsCommit:
+        type: string
+        default: main
+    machine:
+      image: ubuntu-2004:current
```
dmehala, line 83: It may be interesting to update our ci image and use a `docker` executor instead... but it's docker inside docker, it might be problematic.

dgoffredo (approved):
This all looks good to me, per our discussion.

Thanks for adding this.

lgtm

PR #82, opened 2023-12-20
Report `integration` and `integration_version` to telemetry

dmehala:
## Description

This PR send a new type of payload `app_integrations_change` to Telemetry v2. This new payload allows use know which integration is using the Tracer.

### DIFF
--- a/src/datadog/tracer_config.cpp
+++ b/src/datadog/tracer_config.cpp
@@ -366,6 +366,8 @@ Expected<FinalizedTracerConfig> finalize_config(const TracerConfig &config,
   }
 
   result.runtime_id = config.runtime_id;
+  result.integration_name = config.integration_name;
+  result.integration_version = config.integration_version;
 
   return result;
 }
--- a/src/datadog/tracer_config.h
+++ b/src/datadog/tracer_config.h
@@ -119,6 +119,10 @@ struct TracerConfig {
   // such as those in the worker threads/processes of a reverse proxy, might
   // specify the same `runtime_id` for all tracer instances in the same run.
   Optional<RuntimeID> runtime_id;
+
+  // TODO
+  std::string integration_name;
+  std::string integration_version;
 };
 
 // `FinalizedTracerConfig` contains `Tracer` implementation details derived from
@@ -150,6 +154,8 @@ class FinalizedTracerConfig {
   bool report_telemetry;
   Optional<RuntimeID> runtime_id;
   Clock clock;
+  std::string integration_name;
+  std::string integration_version;
 };
 
 // Return a `FinalizedTracerConfig` from the specified `config` and from any
--- a/src/datadog/tracer_config.h
+++ b/src/datadog/tracer_config.h
@@ -121,11 +121,11 @@ struct TracerConfig {
   Optional<RuntimeID> runtime_id;
 
   // `integration_name` is the name of the product integrating this library.
-  // Example: `nginx`, `envoy` or `istio`.
+  // Example: "nginx", "envoy" or "istio".
   std::string integration_name;
   // `integration_version` is the version of the product integrating this
   // library.
-  // Example: `1.2.3`, `6c44da20`, `2020.02.13`
+  // Example: "1.2.3", "6c44da20", "2020.02.13"
   std::string integration_version;
 };
 
--- a/src/datadog/tracer_config.h
+++ b/src/datadog/tracer_config.h
@@ -120,8 +120,12 @@ struct TracerConfig {
   // specify the same `runtime_id` for all tracer instances in the same run.
   Optional<RuntimeID> runtime_id;
 
-  // TODO
+  // `integration_name` is the name of the product integrating this library.
+  // Example: `nginx`, `envoy` or `istio`.
   std::string integration_name;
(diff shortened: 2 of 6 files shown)


dgoffredo:
Each of our "integrations" will need separate consideration for how the designation will make it into dd-trace-cpp.

Here are the ones that come to mind:

- `nginx-datadog` we can do what we want, since we control the code. A `TracerConfig` field will work just fine.
- `envoy` we also have programmatic control over, so we can use `TracerConfig`, but we have to think of how to handle
- `istio`, which uses envoy via a special envoy build called [istio/proxy](https://github.com/istio/proxy), which then gets rolled into images and mentioned in kubernetes YAML configs by [istio](https://github.com/istio/istio) proper. Not sure how best to get the "istio-ness" into dd-trace-cpp, since Istio isn't handling the `TracerConfig` like envoy. Perhaps it's best to add an optional configuration parameter to Envoy's YAML, and have Istio setups default the value when Datadog tracing is mentioned. Tricky.
- `ingress-nginx`, which previously used dd-opentracing-cpp but now is OpenTelemetry-over-Datadog-Agent-only, until I finish the "OpenTelemetry API" projects. If a customer is using the OpenTelemetry way, then we don't have to worry about "integration," because dd-trace-cpp is not involved. If a customer is using a version with dd-opentracing-cpp, then same thing. The question is how `ingress-nginx` will tell its OpenTelemetry nginx module that the Datadog implementation ought to designate itself as `ingress-nginx` :thinking: . That's a tough one.
- Your Apache POC can use `TracerConfig`, no problem.
- Kong doesn't use dd-trace-cpp.

Maybe an environment variable. Doesn't feel right, though.

Have a think about it.

dmehala:
> Maybe an environment variable. Doesn't feel right, though.

Agree, it's a nice workaround but it does not feel appropriate to use an environment environment. It raises concerns about relying on a value that can be changed arbitrarily and depending on users to set it accurately.

Back to our integration challenges, the underlying issue is essentially about propagating the value from X to `dd-trace-cpp`. As a former colleague used to say, [there are no problems we can't solve without another level of indirection.](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering)

### istio integration
I had a quick look on the relationship between `envoy` and `istio`. I started my journey on a naive thought, at some point `istio` needs to inform `envoy` which tracer to use. 

I stumbled upon [istio/pilot/pkg/networking/core/v1alpha3/tracing.go](https://github.com/istio/istio/blob/4718a6ea202e99db516f3953b0ac83fde941b3eb/pilot/pkg/networking/core/v1alpha3/tracing.go#L170-L179), it can return a [datadogConfig](https://github.com/envoyproxy/envoy/blob/6273573560def8329a32584016b6884c7924dc5c/api/envoy/config/trace/v2/datadog.proto#L18). This `DatadogConfig` is then used to set up our Tracer.

A potential solution might involve adding a field to `datadog.proto` and seting that field in `istio` code.

### ingress-nginx
No idea 🤷‍♂️ Further investigation is required.

src/datadog/tracer_config.cpp
```diff
@@ -366,6 +366,7 @@ Expected<FinalizedTracerConfig> finalize_config(const TracerConfig &config,
   }
 
   result.runtime_id = config.runtime_id;
+  result.integration_name = config.integration_name;
```
dgoffredo, line 369: Where does "integration _version_" come from? Here we are in the land of names...

- Tracing library is dd-trace-cpp and its release version.
- Service is designated by the end-user (but the integration chooses a default). `config.defaults.version` is the version of that service, so also determined by the end-user.
- Integration is determined by who is offering a trace-compatible product to the end-user. So that might be the maintainers of Envoy, Istio, nginx-datadog, Kong, etc. The integration has a version separate from the other two notions of version. Let's call it the "integration version."

Where does "integration version" get configured? I see that it ends up in `class Tracer`.

dgoffredo (commented):
As we discussed in a recent meeting, once you figure out whether there's a difference between:

- the absence of the integration fields
- the presence of the integration fields, but as empty strings

then we can finalize the policy about `TracerConfig::integration_*`, document them, and be done with this.

The code and test look good.

src/datadog/tracer_config.h
```diff
@@ -119,6 +119,14 @@ struct TracerConfig {
...
   // specify the same `runtime_id` for all tracer instances in the same run.
   Optional<RuntimeID> runtime_id;
+
+  // `integration_name` is the name of the product integrating this library.
+  // Example: `nginx`, `envoy` or `istio`.
+  std::string integration_name;
+  // `integration_version` is the version of the product integrating this
+  // library.
+  // Example: `1.2.3`, `6c44da20`, `2020.02.13`
+  std::string integration_version;
```
dgoffredo, line 129:
```suggestion
  // `integration_name` is the name of the product integrating this library.
  // Example: "nginx", "envoy" or "istio".
  std::string integration_name;
  // `integration_version` is the version of the product integrating this
  // library.
  // Example: "1.2.3", "6c44da20", "2020.02.13"
  std::string integration_version;
```

dgoffredo (approved):
Nicely done.

PR #86, opened 2024-01-08
Configure parametric test to get the cpp tracer in same way as system-tests

robertomonteromiguel:
Configure parametric test to get the cpp tracer in same way as system-tests

Related with system-tests PR: https://github.com/DataDog/system-tests/pull/1983

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,7 +95,7 @@ jobs:
             cd system-tests
             git remote add origin https://github.com/DataDog/system-tests.git
             #git fetch origin << parameters.systemTestsCommit >>
-            git fetch origin 5598dcb607258ad2ee5be79917a344197f862c28
+            git fetch origin f9b90d5ff2ecc7937e68d572ae3c8dcc6812234b
             git reset --hard FETCH_HEAD
       - run:
           name: Install requirements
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -105,12 +105,15 @@ jobs:
...
+
       - run:
-          name: Run
+          name: Run Parametric tests
           environment:
             TEST_LIBRARY: cpp
             DD_TRACE_CPP_COMMIT: << pipeline.git.revision >>
-          command: cd system-tests && ./build.sh -i runner && ./run.sh PARAMETRIC --log-cli-level=DEBUG 
+          command: |
+            echo "https://github.com/DataDog/dd-trace-cpp@$DD_TRACE_CPP_COMMIT" > ~/project/system-tests/binaries/cpp-load-from-git
```
dmehala, line 115: Does this actually work?

robertomonteromiguel: If it will work once we merge the PR: https://github.com/DataDog/system-tests/pull/1983

dmehala (approved):
LGTM :shipit:

.circleci/config.yml
dmehala: Could you just set the version of `dd-trace-cpp` instead of the whole github url?

robertomonteromiguel: It is to maintain a certain similarity with other tracers. ie: https://github.com/DataDog/system-tests/blob/main/utils/scripts/load-binary.sh#L230

PR #87, opened 2024-01-08
Add windows support

dmehala:


dmehala:
That benchmark is... not stable. First run was close to 5% and now 4.1%, while on my machine this is a big gap (I think, need to check again).

dmehala:
Soon™️

PR #89, opened 2024-01-12
TraceSegment can outlive Tracer

dgoffredo:
I was integrating the recent sampling delegation changes into the corresponding feature branch in nginx-datadog, and got some compiler errors. They're due to other changes on the dd-trace-cpp main branch.

In nginx-datadog, and in Envoy, the `Tracer` is stored as an `Optional<Tracer>`. A `Tracer` instance is then moved into the variable.

Since `Tracer` now has a `ConfigManager` data member, and `ConfigManager` has a `std::mutex` data member, `Tracer` can no longer be moved (because `std::mutex` cannot be moved).

One thing I could do is use `Optional::emplace` to construct the `Tracer` in place: no move required.

Another thing I could do is make the `Tracer`'s `ConfigManager` a `std::unique_ptr`, so that move semantics are preserved.

Yet another thing I could do is use a `std::unique_ptr` in nginx-datadog and Envoy, so that the `Tracer` itself does not have to be moved.

So, I went about thinking about this. Then, looking at the dd-trace-cpp code, I noticed something.

`Tracer` has `ConfigManager` data member. A reference (pointer) to the `ConfigManager` is then passed into the `DatadogAgent`, which is held by `shared_ptr` in `Tracer`. The `DatadogAgent` has a `RemoveConfigurationManager` data member, which holds a reference (pointer) to the `ConfigManager`.

My initial design of dd-trace-cpp was such that a `TraceSegment` could safely outlive the `Tracer` that created it. The idea was that an application might have a `Tracer` that gets reassigned when reconfiguration occurs, and this would not affect any traces in flight.

The recent remote configuration changes violate this property, because the lifetime of the `ConfigManager` is limited to the lifetime of the `Tracer` that holds it, even though the `ConfigManager` is referred to by objects that can outlive the `Tracer`.

This pull request makes the `Tracer`'s `ConfigManager` a `std::shared_ptr`, and passes it as a `shared_ptr` through `DatadogAgent` and into `RemoteConfigurationManager`. This way, even if the `Tracer` is destroyed, the `RemoteConfigurationManager` still has a valid handle to the `ConfigManager`.

### DIFF
--- a/src/datadog/remote_config.h
+++ b/src/datadog/remote_config.h
@@ -12,6 +12,7 @@
 // It interacts with the `ConfigManager` to seamlessly apply or revert
 // configurations based on responses received from the remote source.
 
+#include <memory>
 #include <string>
 
 #include "config_manager.h"
@@ -44,15 +45,16 @@ class RemoteConfigurationManager {
   };
 
   TracerSignature tracer_signature_;
-  ConfigManager& config_manager_;
+  std::shared_ptr<ConfigManager> config_manager_;
   std::string client_id_;  ///< Identifier a `RemoteConfigurationManager`
 
   State state_;
   std::unordered_map<std::string, Configuration> applied_config_;
 
  public:
-  RemoteConfigurationManager(const TracerSignature& tracer_signature,
-                             ConfigManager& config_manager);
+  RemoteConfigurationManager(
+      const TracerSignature& tracer_signature,
+      const std::shared_ptr<ConfigManager>& config_manager);
 
   // Construct a JSON object representing the payload to be sent in a remote
   // configuration request.
--- a/src/datadog/remote_config.h
+++ b/src/datadog/remote_config.h
@@ -46,7 +46,7 @@ class RemoteConfigurationManager {
 
   TracerSignature tracer_signature_;
   std::shared_ptr<ConfigManager> config_manager_;
-  std::string client_id_;  ///< Identifier a `RemoteConfigurationManager`
+  std::string client_id_;
 
   State state_;
   std::unordered_map<std::string, Configuration> applied_config_;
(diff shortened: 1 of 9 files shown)


src/datadog/tracer.cpp
```diff
@@ -36,6 +36,7 @@ Tracer::Tracer(const FinalizedTracerConfig& config)
 Tracer::Tracer(const FinalizedTracerConfig& config,
                const std::shared_ptr<const IDGenerator>& generator)
     : logger_(config.logger),
+      config_manager_(std::make_shared<ConfigManager>(config)),
```
dgoffredo, line 39: I moved this up so that it occurs before `collector_`. This is not at all necessary, since we're talking about shared pointers, but it mirrors what we would do if the collector and the config manager were both held by value.

src/datadog/remote_config.h
```diff
@@ -44,15 +45,16 @@ class RemoteConfigurationManager {
   };
 
   TracerSignature tracer_signature_;
-  ConfigManager& config_manager_;
-  std::string client_id_;  ///< Identifier a `RemoteConfigurationManager`
```
dmehala, line 48: ?

src/datadog/tracer.h
```diff
@@ -10,6 +10,9 @@
 // obtained from a `TracerConfig` via the `finalize_config` function.  See
 // `tracer_config.h`.
 
+#include <cstddef>
```
dmehala, line 13: no needed

dmehala (approved):

src/datadog/remote_config.h
dgoffredo: I figured the comment contained a typo:
```
///< Identifier a `RemoteConfigurationManager`
```
I'm not sure what the extra `/>` is doing, and "Identifier" should probably be "Identifies".

Ok, suppose the comment is "Identifies a `RemoteConfigurationManager`." True enough, but that doesn't really help me understand what `client_id_` is or is for. So, I thought about how best to describe `client_id_` in a comment. I didn't think of anything that was better than nothing, so I went with nothing.

dgoffredo:
Thanks for the review.

I'll log off after this merge. It's a USA holiday today :D

PR #93, opened 2024-01-25
chore-ci: install python3

dmehala:
# Description
Adds `python` to our CI image.

## Why?
I recently enforced formating/linting in `nginx-datadog`. They share the same CI image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,4 +36,3 @@ run chmod +x /tmp/install-cmake && /tmp/install-cmake && rm /tmp/install-cmake
 copy bin/install-lcov /tmp/install-lcov
 run chmod +x /tmp/install-lcov && /tmp/install-lcov && rm /tmp/install-lcov
 
-run apt-get clean -y
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ run apt-get update && apt-get install --yes software-properties-common && \
     apt-get update && apt-get upgrade --yes && \
     apt-get install --yes \
         wget build-essential clang sed gdb clang-format git ssh shellcheck \
-        libc++-dev libc++abi-dev python3
+        libc++-dev libc++abi-dev python3 pip
 
 # bazelisk, a launcher for bazel. `bazelisk --help` will cause the latest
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -35,3 +35,5 @@ run chmod +x /tmp/install-cmake && /tmp/install-cmake && rm /tmp/install-cmake
 # Coverage reporting.
 copy bin/install-lcov /tmp/install-lcov
 run chmod +x /tmp/install-lcov && /tmp/install-lcov && rm /tmp/install-lcov
+
+run apt-get clean -y
```
dgoffredo, line 39: This had me googling. It's new to me.

Have a look:

- https://gist.github.com/marvell/7c812736565928e602c4
- https://stackoverflow.com/questions/61990329/benefits-of-repeated-apt-cache-cleans
- https://docs.docker.com/develop/develop-images/instructions/#apt-get

I recommend we omit the call to `apt-get clean`, as well as omit any `rm`. I doubt it saves much space, and it's not clear which incantation is "best."

Dockerfile
```diff
@@ -20,7 +20,7 @@ run apt-get update && apt-get install --yes software-properties-common && \
     apt-get update && apt-get upgrade --yes && \
     apt-get install --yes \
         wget build-essential clang sed gdb clang-format git ssh shellcheck \
-        libc++-dev libc++abi-dev
+        libc++-dev libc++abi-dev python3 pip
```
dgoffredo, line 23: In addition to, or as an alternative to, `pip`, we could install the `yapf3` package. Then, in nginx-datadog's CI, we wouldn't need the [install python dependencies](https://github.com/DataDog/nginx-datadog/blob/6b36bb93287963479ca3cfe73e24a51846f7cd66/.circleci/config.yml#L127-L130) step.

Dockerfile
dmehala: ok

dmehala: Fixed in 8769e8cf7c359e45f8cdbb8a80a36bc8a443634f

PR #95, opened 2024-01-29
fix: remote configuration default targets version

dmehala:
# Description
The RFC reference that the default value of `targets_version` should be `0` and not `1`. This PR fix this mistake and also resolves an issue where the agent was producing an error in some situation.

### DIFF
--- a/src/datadog/remote_config.h
+++ b/src/datadog/remote_config.h
@@ -30,7 +30,7 @@ class RemoteConfigurationManager {
   // Represents the *current* state of the RemoteConfigurationManager.
   // It is also used to report errors to the remote source.
   struct State {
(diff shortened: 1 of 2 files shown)


dgoffredo:
> and also resolves an issue where the agent was producing an error in some situation

Was that the scenario when the Agent was returning 200 OK with an error message derived from an "unknown" gRPC error from the backend?

dgoffredo (approved):
Nice work, thanks for the fix.

This brings up an optional "TODO" that one of us can address later: submit a PR to datadog-agent to ensure that non-JSON responses always have non-2xx statuses. I think it's a little logic error in their gRPC status `switch` logic that allowed the Agent to return 200 with an error message (instead of JSON).

PR #96, opened 2024-01-29
feat: support DD_TRACE_ENABLED dynamic configuration

dmehala:


src/datadog/config_manager.h
```diff
@@ -34,6 +37,8 @@ class ConfigManager {
   // Return the `SpanDefaults` consistent with the most recent configuration.
   std::shared_ptr<const SpanDefaults> get_span_defaults();
 
+  bool get_report_traces();
```
dmehala, line 40: - [ ] document (and rename?)

dmehala (commented):
**TODO:**
- [x] write test
- [ ] ~~remove `null_collector.*`~~ (could be useful)

src/datadog/config_manager.h
dgoffredo: We could follow the lead of `TracerConfig` and omit the `get_` prefix. It's fine either way.

Yes, please do add a comment here. Maybe:
```
// Return whether spans should be sent to the collector.
```
I don't like the use of the word "should." Something like that, anyway.

src/datadog/remote_config.cpp
```diff
@@ -94,6 +96,15 @@ ConfigUpdate parse_dynamic_config(const nlohmann::json& j) {
     }
   }
 
+  if (auto tracing_enabled_it = j.find("tracing_enabled");
+      tracing_enabled_it != j.cend()) {
+    if (tracing_enabled_it->is_boolean()) {
+      config_update.report_traces = tracing_enabled_it->get<bool>();
+    } else {
+      // TODO: report to telemetry
```
dgoffredo, line 104: coming soon...

dgoffredo (approved):
:shipit:

src/datadog/remote_config.cpp
dmehala: image

PR #98, opened 2024-02-09
[2/2] feat: active configuration - report configuration to telemetry

dmehala:
# Description
This PR is second and final part of active configuration. This particular PR report configurations in `app-started` event and also report configuration updates received through Remote Configuration.

### TODO
- [x] Write unit tests.
- [ ] Validate with system-tests.

### DIFF
--- a/src/datadog/config_manager.cpp
+++ b/src/datadog/config_manager.cpp
@@ -3,34 +3,11 @@
 #include <sstream>
 
 #include "parse_util.h"
+#include "string_util.h"
 #include "trace_sampler.h"
 
 namespace datadog {
 namespace tracing {
-namespace {
-
-std::string join(const std::vector<StringView>& values,
-                 const char* const separator) {
-  if (values.empty()) return {};
-  auto it = values.cbegin();
-
-  std::string res{*it};
-  for (++it; it != values.cend(); ++it) {
-    res += separator;
-    append(res, *it);
-  }
-
-  return res;
-}
-
-// TODO: use `to_chars`
-std::string to_string(double d) {
-  std::stringstream stream;
-  stream << std::fixed << std::setprecision(1) << d;
-  return stream.str();
-}
-
-}  // namespace
 
 ConfigManager::ConfigManager(const FinalizedTracerConfig& config)
     : clock_(config.clock),
@@ -76,16 +53,15 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
     auto trace_sampler =
         std::make_shared<TraceSampler>(*finalized_trace_sampler_cfg, clock_);
 
-    if (trace_sampler != trace_sampler_.get()) {
-      trace_sampler_ = std::move(trace_sampler);
-      telemetry_conf.emplace_back(std::move(trace_sampling_metadata));
-    }
-  } else {
-    if (!trace_sampler_.is_default()) {
-      trace_sampler_.reset();
-      telemetry_conf.emplace_back(
-          default_config_metadata_[ConfigName::TRACE_SAMPLING_RATE]);
-    }
+    // This reset rate limiting and `TraceSampler` has no `operator==`.
+    // TODO: Instead of creating another `TraceSampler`, we should
+    // update the default sampling rate
+    trace_sampler_ = std::move(trace_sampler);
+    telemetry_conf.emplace_back(std::move(trace_sampling_metadata));
+  } else if (!trace_sampler_.is_original_value()) {
+    trace_sampler_.reset();
+    telemetry_conf.emplace_back(
+        default_config_metadata_[ConfigName::TRACE_SAMPLING_RATE]);
   }
 
   if (conf.tags) {
@@ -105,26 +81,22 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
       span_defaults_ = new_span_defaults;
       telemetry_conf.emplace_back(std::move(tags_metadata));
     }
-  } else {
-    if (!span_defaults_.is_default()) {
-      span_defaults_.reset();
-      telemetry_conf.emplace_back(default_config_metadata_[ConfigName::TAGS]);
-    }
+  } else if (!span_defaults_.is_original_value()) {
+    span_defaults_.reset();
+    telemetry_conf.emplace_back(default_config_metadata_[ConfigName::TAGS]);
   }
 
   if (conf.report_traces) {
     if (conf.report_traces != report_traces_) {
       report_traces_ = *conf.report_traces;
       telemetry_conf.emplace_back(ConfigName::REPORT_TRACES,
-                                  *conf.report_traces ? "true" : "false",
+                                  to_string(*conf.report_traces),
                                   ConfigMetadata::Origin::REMOTE_CONFIG);
     }
-  } else {
-    if (!report_traces_.is_default()) {
-      report_traces_.reset();
-      telemetry_conf.emplace_back(
-          default_config_metadata_[ConfigName::REPORT_TRACES]);
-    }
+  } else if (!report_traces_.is_original_value()) {
+    report_traces_.reset();
+    telemetry_conf.emplace_back(
+        default_config_metadata_[ConfigName::REPORT_TRACES]);
   }
 
   return telemetry_conf;
@@ -134,18 +106,18 @@ std::vector<ConfigMetadata> ConfigManager::reset() {
   std::vector<ConfigMetadata> config_metadata;
 
   std::lock_guard<std::mutex> lock(mutex_);
-  if (!trace_sampler_.is_default()) {
+  if (!trace_sampler_.is_original_value()) {
     trace_sampler_.reset();
     config_metadata.emplace_back(
         default_config_metadata_[ConfigName::TRACE_SAMPLING_RATE]);
   }
 
-  if (!span_defaults_.is_default()) {
+  if (!span_defaults_.is_original_value()) {
     span_defaults_.reset();
     config_metadata.emplace_back(default_config_metadata_[ConfigName::TAGS]);
   }
 
-  if (!report_traces_.is_default()) {
+  if (!report_traces_.is_original_value()) {
     report_traces_.reset();
     config_metadata.emplace_back(
         default_config_metadata_[ConfigName::REPORT_TRACES]);
--- a/src/datadog/config_manager.cpp
+++ b/src/datadog/config_manager.cpp
@@ -72,12 +72,15 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
     auto finalized_trace_sampler_cfg = finalize_config(trace_sampler_cfg);
     if (auto error = finalized_trace_sampler_cfg.if_error()) {
       trace_sampling_metadata.error = *error;
-    } else {
-      trace_sampler_ =
-          std::make_shared<TraceSampler>(*finalized_trace_sampler_cfg, clock_);
     }
 
-    telemetry_conf.emplace_back(std::move(trace_sampling_metadata));
+    auto trace_sampler =
+        std::make_shared<TraceSampler>(*finalized_trace_sampler_cfg, clock_);
+
+    if (trace_sampler != trace_sampler_.get()) {
+      trace_sampler_ = std::move(trace_sampler);
+      telemetry_conf.emplace_back(std::move(trace_sampling_metadata));
+    }
   } else {
     if (!trace_sampler_.is_default()) {
       trace_sampler_.reset();
@@ -93,15 +96,16 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
     auto parsed_tags = parse_tags(*conf.tags);
     if (auto error = parsed_tags.if_error()) {
       tags_metadata.error = *error;
-    } else {
+    }
+
+    if (*parsed_tags != span_defaults_.get()->tags) {
       auto new_span_defaults =
           std::make_shared<SpanDefaults>(*span_defaults_.get());
       new_span_defaults->tags = std::move(*parsed_tags);
 
       span_defaults_ = new_span_defaults;
+      telemetry_conf.emplace_back(std::move(tags_metadata));
     }
-
-    telemetry_conf.emplace_back(std::move(tags_metadata));
   } else {
     if (!span_defaults_.is_default()) {
       span_defaults_.reset();
@@ -110,10 +114,12 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
   }
 
   if (conf.report_traces) {
-    ConfigMetadata tags_metadata(ConfigName::REPORT_TRACES,
-                                 *conf.report_traces ? "true" : "false",
-                                 ConfigMetadata::Origin::REMOTE_CONFIG);
-    report_traces_ = *conf.report_traces;
+    if (conf.report_traces != report_traces_) {
+      report_traces_ = *conf.report_traces;
+      telemetry_conf.emplace_back(ConfigName::REPORT_TRACES,
+                                  *conf.report_traces ? "true" : "false",
+                                  ConfigMetadata::Origin::REMOTE_CONFIG);
+    }
   } else {
     if (!report_traces_.is_default()) {
       report_traces_.reset();
--- a/src/datadog/config_manager.cpp
+++ b/src/datadog/config_manager.cpp
@@ -11,7 +11,7 @@ namespace tracing {
 
 ConfigManager::ConfigManager(const FinalizedTracerConfig& config)
     : clock_(config.clock),
-      default_config_metadata_(config.metadata),
+      default_metadata_(config.metadata),
       trace_sampler_(
           std::make_shared<TraceSampler>(config.trace_sampler, clock_)),
       span_defaults_(std::make_shared<SpanDefaults>(config.defaults)),
@@ -19,27 +19,30 @@ ConfigManager::ConfigManager(const FinalizedTracerConfig& config)
 
 std::shared_ptr<TraceSampler> ConfigManager::trace_sampler() {
   std::lock_guard<std::mutex> lock(mutex_);
-  return trace_sampler_;
+  return trace_sampler_.value();
 }
 
 std::shared_ptr<const SpanDefaults> ConfigManager::span_defaults() {
   std::lock_guard<std::mutex> lock(mutex_);
-  return span_defaults_;
+  return span_defaults_.value();
 }
 
 bool ConfigManager::report_traces() {
   std::lock_guard<std::mutex> lock(mutex_);
-  return report_traces_;
+  return report_traces_.value();
 }
 
 std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
-  std::vector<ConfigMetadata> telemetry_conf;
+  std::vector<ConfigMetadata> metadata;
 
   std::lock_guard<std::mutex> lock(mutex_);
 
-  if (conf.trace_sampling_rate) {
+  if (!conf.trace_sampling_rate) {
+    reset_config(ConfigName::TRACE_SAMPLING_RATE, trace_sampler_, metadata);
+  } else {
     ConfigMetadata trace_sampling_metadata(
-        ConfigName::TRACE_SAMPLING_RATE, to_string(*conf.trace_sampling_rate),
+        ConfigName::TRACE_SAMPLING_RATE,
+        to_string(*conf.trace_sampling_rate, 1),
         ConfigMetadata::Origin::REMOTE_CONFIG);
 
     TraceSamplerConfig trace_sampler_cfg;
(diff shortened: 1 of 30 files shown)


src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,159 @@
...
+
+  return res;
+}
+
+// TODO: use `to_chars`
+std::string to_string(double d) {
+  std::stringstream stream;
+  stream << std::fixed << std::setprecision(1) << d;
+  return stream.str();
+}
```
dmehala, line 31: - [ ] Use `to_chars`

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,159 @@
...
-      current_trace_sampler_ =
-          std::make_shared<TraceSampler>(*finalized_trace_sampler_cfg, clock_);
+  if (conf.trace_sampling_rate) {
+    ConfigMetadata trace_sampling_metadata(
+        ConfigName::TRACE_SAMPLING_RATE, to_string(*conf.trace_sampling_rate),
+        ConfigMetadata::Origin::REMOTE_CONFIG);
+
+    TraceSamplerConfig trace_sampler_cfg;
+    trace_sampler_cfg.sample_rate = *conf.trace_sampling_rate;
+    trace_sampler_cfg.max_per_second = 200;
```
dmehala, line 70: - [ ] Remove. This should be addressed by `finalize_config` instead.

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,159 @@
...
+    if (!span_defaults_.is_default()) {
+      span_defaults_.reset();
+      telemetry_conf.emplace_back(default_config_metadata_[ConfigName::TAGS]);
+    }
   }
 
   if (conf.report_traces) {
-    current_report_traces_ = *conf.report_traces;
+    ConfigMetadata tags_metadata(ConfigName::REPORT_TRACES,
+                                 *conf.report_traces ? "true" : "false",
```
dmehala, line 114: - [ ] Use `to_string(bool)` instead.

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,159 @@
...
+                                 ConfigMetadata::Origin::REMOTE_CONFIG);
+    report_traces_ = *conf.report_traces;
   } else {
-    current_report_traces_ = default_report_traces_;
+    if (!report_traces_.is_default()) {
+      report_traces_.reset();
+      telemetry_conf.emplace_back(
+          default_config_metadata_[ConfigName::REPORT_TRACES]);
+    }
```
dmehala, line 122: Redundant. Maybe create `reset_X` private method, use it here and in `reset()`?

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,159 @@
...
 }
 
 nlohmann::json ConfigManager::config_json() const {
   std::lock_guard<std::mutex> lock(mutex_);
-  return nlohmann::json{
-      {"default", to_json(*current_span_defaults_)},
-      {"trace_sampler", current_trace_sampler_->config_json()},
-      {"report_traces", current_report_traces_}};
+  return nlohmann::json{{"default", to_json(*span_defaults_.get())},
+                        {"trace_sampler", trace_sampler_.get()->config_json()},
```
dmehala, line 155: Ugly.

src/datadog/config_update.h
```diff
@@ -15,8 +15,8 @@ namespace tracing {
 // remote configuration value.
 struct ConfigUpdate {
   Optional<bool> report_traces;
-  Optional<TraceSamplerConfig> trace_sampler;
-  Optional<std::unordered_map<std::string, std::string>> tags;
+  Optional<double> trace_sampling_rate;
```
dmehala, line 18: Maybe use `Rate` instead of `double`.

src/datadog/tracer_config.cpp
```diff
@@ -742,38 +848,46 @@ Expected<FinalizedTracerConfig> finalize_config(
...
+    origin = ConfigMetadata::Origin::ENVIRONMENT_VARIABLE;
   } else if (user_config.trace_sampler.sample_rate) {
     ts.sample_rate = user_config.trace_sampler.sample_rate;
+    origin = ConfigMetadata::Origin::CODE;
   }
 
+  if (ts.sample_rate) {
+    final_cfg.metadata[ConfigName::TRACE_SAMPLING_RATE] = ConfigMetadata(
+        ConfigName::TRACE_SAMPLING_RATE, to_string(*ts.sample_rate), origin);
+  }
```
dmehala, line 860:
```suggestion
    final_cfg.metadata[ConfigName::TRACE_SAMPLING_RATE] = ConfigMetadata(
        ConfigName::TRACE_SAMPLING_RATE, to_string(ts.sample_rate.value_or(1.0)), origin);
```

src/datadog/tracer_config.h
```diff
@@ -176,6 +178,79 @@ struct TracerDefaultConfig final {
...
+    case ConfigName::TRACE_SAMPLING_LIMIT:
+      return "DD_TRACE_RATE_LIMIT";
+    case ConfigName::SPAN_SAMPLING_RULES:
+      return "DD_SPAN_SAMPLING_RULES";
+    case ConfigName::TRACE_SAMPLING_RULES:
+      return "DD_TRACE_SAMPLING_RULES";
+  }
+
+  std::abort();
+}
```
dmehala, line 240: that's telemetry name, maybe move this conversion method to `tracer_telemetry.cpp`?

src/datadog/tracer_telemetry.h
```diff
@@ -49,6 +50,8 @@ class TracerTelemetry {
   std::string integration_name_;
   std::string integration_version_;
   uint64_t seq_id_ = 0;
```
dmehala, line 52: - [ ] Add comment: `// Track sequence id per payload generated`

src/datadog/datadog_agent.cpp
```diff
@@ -433,10 +426,11 @@ void DatadogAgent::get_and_apply_remote_configuration_updates() {
...
         if (!response_json.empty()) {
-          // TODO (during Active Configuration): `process_response` should
-          // return a list of configuration update and should be consumed by
-          // telemetry.
-          remote_config_.process_response(response_json);
+          auto updated_configuration =
+              remote_config_.process_response(response_json);
+          if (!updated_configuration.empty()) {
+            send_configuration_change(updated_configuration);
+          }
```
dgoffredo, line 433: You might run into a bug that I fixed in an unmerged [ancient branch][1].

We're currently in a callback (response handler) to `class Curl`. `Curl`'s mutex is locked, so when we try to enqueue a HTTP request inside `send_configuration_change`, I think we'll have a deadlock.

Easy enough to fix `class Curl`.

[1]: https://github.com/DataDog/dd-trace-cpp/compare/main...david.goffredo/curl-event-loop#diff-a747ca63fc60398d29a5669c3a753950c18e737d84cbd518fca99f5752bd60abR443-R458

src/datadog/tracer_telemetry.cpp
```diff
@@ -241,5 +332,22 @@ std::string TracerTelemetry::app_closing() {
...
+    const std::vector<ConfigMetadata>& new_configuration) {
+  auto configuration_json = nlohmann::json::array();
+  for (const auto& config_metadata : new_configuration) {
+    // if (config_metadata.value.empty()) continue;
+    configuration_json.emplace_back(
+        generate_configuration_field(config_metadata));
+  }
+
+  auto configuration_change =
+      generate_telemetry_body("app-client-configuration-change");
```
dgoffredo, line 345: Be sure to mention this in the comments at the top of `tracer_telemetry.h`. That is also a good time to audit the code for any telemetry event types that we forgot to mention before.

dgoffredo (commented):
Here is a first pass review for our meeting today.

My thoughts so far:

- Let's consider batching configuration updates, sending them only whenever the telemetry timer fires.
- Let's think more about how `ConfigManager` stores and modifies its state. I need to spend more time looking at it.
- Take a look at the possible deadlock that I commented on inline.
- How are we going to test this? I'll have to look more at our existing tests.

src/datadog/datadog_agent.cpp
dgoffredo: Nope, we fixed it a while back.

src/datadog/tracer_config.cpp
```diff
@@ -742,38 +848,45 @@ Expected<FinalizedTracerConfig> finalize_config(
...
   } else if (user_config.trace_sampler.sample_rate) {
     ts.sample_rate = user_config.trace_sampler.sample_rate;
+    origin = ConfigMetadata::Origin::CODE;
   }
 
+  final_cfg.metadata[ConfigName::TRACE_SAMPLING_RATE] =
+      ConfigMetadata(ConfigName::TRACE_SAMPLING_RATE,
+                     to_string(ts.sample_rate.value_or(1.0)), origin);
+
+  // TS: Rules
```
dgoffredo, line 861: I'd prefer a longer variable name than `ts`. But, if you are going with `ts`, then at least use the full name in comments.
```c++
// trace sampler rules
...
// trace sampler max per second
```

src/datadog/tracer_config.cpp
```diff
@@ -742,38 +848,45 @@ Expected<FinalizedTracerConfig> finalize_config(
...
-                               user_config.trace_sampler.max_per_second, 200);
+  // TS: Max Per Second
+  std::tie(origin, ts.max_per_second) =
+      pick(env->trace_sampler.max_per_second,
+           user_config.trace_sampler.max_per_second, 200);
+  final_cfg.metadata[ConfigName::TRACE_SAMPLING_LIMIT] = ConfigMetadata(
+      ConfigName::TRACE_SAMPLING_LIMIT, to_string(*ts.max_per_second), origin);
 
   SpanSamplerConfig ss;
+  // SS: Rules
```
dgoffredo, line 880: [ss](https://man7.org/linux/man-pages/man8/ss.8.html), [ss](https://en.wikipedia.org/wiki/Schutzstaffel), [ss](https://en.wikipedia.org/wiki/SS_(disambiguation))

src/datadog/tracer_config.h
```diff
@@ -13,6 +13,7 @@
 #include "datadog_agent_config.h"
 #include "error.h"
 #include "expected.h"
+#include "json.hpp"
```
dgoffredo, line 16: Is `nlohmann::json` used in this header? If not, remove this include. If so, include `json_fwd.hpp` instead.

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,164 @@
...
+
+  return res;
+}
+
+// TODO: use `to_chars`
+std::string to_string(double d) {
+  std::stringstream stream;
+  stream << std::fixed << std::setprecision(1) << d;
+  return stream.str();
+}
```
dgoffredo, line 31: This code is similar to that in `tracer_config.cpp`. Perhaps all of these should be in a dedicated component. One advantage would be knowing there's one place to look for `to_string` and `join` overloads, as opposed to wondering "is it a local utility, or declared in some header I included?"

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,164 @@
...
+
+    auto finalized_trace_sampler_cfg = finalize_config(trace_sampler_cfg);
+    if (auto error = finalized_trace_sampler_cfg.if_error()) {
+      trace_sampling_metadata.error = *error;
+    }
+
+    auto trace_sampler =
+        std::make_shared<TraceSampler>(*finalized_trace_sampler_cfg, clock_);
+
+    if (trace_sampler != trace_sampler_.get()) {
```
dgoffredo, line 79: This is comparing `shared_ptr`s.

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,164 @@
...
+      trace_sampler_ = std::move(trace_sampler);
+      telemetry_conf.emplace_back(std::move(trace_sampling_metadata));
     }
   } else {
-    current_trace_sampler_ = default_trace_sampler_;
+    if (!trace_sampler_.is_default()) {
```
dgoffredo, line 84: For our meeting: What is the meaning of "default" in this context?

src/datadog/config_manager.cpp
```diff
@@ -1,80 +1,164 @@
...
+      auto new_span_defaults =
+          std::make_shared<SpanDefaults>(*span_defaults_.get());
+      new_span_defaults->tags = std::move(*parsed_tags);
+
+      span_defaults_ = new_span_defaults;
+      telemetry_conf.emplace_back(std::move(tags_metadata));
+    }
   } else {
-    current_span_defaults_ = default_span_defaults_;
+    if (!span_defaults_.is_default()) {
```
dgoffredo, line 109: `else if`?

src/datadog/tracer_telemetry.cpp
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/tracer_config.cpp
```diff
@@ -21,48 +21,86 @@ namespace datadog {
...
+        return "Datadog";
+      case PropagationStyle::W3C:
+        return "W3C";
+      case PropagationStyle::NONE:
+        return "None";
+    }
+    return "";  ///< unlikely
+  };
+
+  if (values.empty()) return {};
```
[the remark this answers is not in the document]
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/tracer_config.cpp
```diff
@@ -648,6 +659,19 @@ Expected<FinalizedTracerConfig> finalize_config(
...
+                                          const Optional<T> &user,
+                                          U default_v) {
+  if (env) {
+    return {ConfigMetadata::Origin::ENVIRONMENT_VARIABLE, *env};
+  } else if (user) {
+    return {ConfigMetadata::Origin::CODE, *user};
+  }
+  return {ConfigMetadata::Origin::DEFAULT, default_v};
+}
+
```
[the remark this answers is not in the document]
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/tracer_config.cpp
```diff
@@ -665,45 +689,113 @@ Expected<FinalizedTracerConfig> finalize_config(
   }
 
   final_cfg.clock = clock;
-  final_cfg.defaults.service = value_or(env->service, user_config.service, "");
+
```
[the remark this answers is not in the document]
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/tracer_config.cpp
```diff
@@ -665,45 +689,113 @@ Expected<FinalizedTracerConfig> finalize_config(
...
+  // Tags Header Size
   final_cfg.tags_header_size =
       value_or(env->tags_header_size, user_config.tags_header_size,
                default_config.max_tags_header_size);
-  final_cfg.trace_id_128_bit =
-      value_or(env->trace_id_128_bit, user_config.trace_id_128_bit,
-               default_config.generate_128bit_trace_ids);
+
+  // 128b Trace Ids
```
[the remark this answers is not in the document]
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/config_manager.cpp
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/config_manager.cpp
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/config_manager.cpp
dmehala: addressed in e2bb8f465b2245055cb150d6d5bc8bcf69211c65

src/datadog/tracer_config.cpp
```diff
@@ -21,48 +21,86 @@ namespace datadog {
 namespace tracing {
 namespace {
 
-bool falsy(StringView text) {
-  auto lower = std::string{text};
-  to_lower(lower);
-  return lower == "0" || lower == "false" || lower == "no";
+std::string to_string(bool b) { return b ? "true" : "false"; }
+
+// TODO: use `to_chars`
```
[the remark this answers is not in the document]
dmehala: To avoid:
Screenshot 2024-03-01 at 11 16 16

src/datadog/string_util.cpp
```diff
@@ -0,0 +1,81 @@
...
+  for (++it; it != values.cend(); ++it) {
+    res += ",";
+    res += it->first;
+    res += ":";
+    res += it->second;
+  }
+
+  return res;
+}
+
```
dgoffredo, line 78: I couldn't resist...
```c++
namespace {

template <typename Sequence, typename Func>
std::string join(const Sequence& elements, StringView separator, Func&& append_element) {
    auto iter = std::begin(elements);
    const auto end = std::end(elements);
    std::string result;
    if (iter == end) {
        return result;
    }
    append_element(result, *iter);
    for (++iter; iter != end; ++iter) {
        append(result, separator);
        append_element(result, *iter);
    }
    return result;
}

} // namespace

std::string join_tags(
    const std::unordered_map<std::string, std::string>& values) {
  return join(values, ",", [](std::string& result, const auto& entry) {
    const auto& [key, value] = entry;
    result += key;
    result += ':';
    result += value;
  });
}

std::string join(const std::vector<StringView>& values,
                 StringView separator) {
    return join(values, separator, [](std::string& result, StringView value) {
        append(result, value);
    });
}

std::string join_propagation_styles(
    const std::vector<PropagationStyle>& values) {
  return join(values, ",", [](std::string& result, PropagationStyle style) {
    switch (style) {
      case PropagationStyle::B3:
        result += "b3"; break;
      case PropagationStyle::DATADOG:
        result += "datadog"; break;
      case PropagationStyle::W3C:
        result += "tracecontext"; break;
      case PropagationStyle::NONE:
        result += "none"; break;
    }
  });
}
```

src/datadog/string_util.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <string>
+
+#include "propagation_style.h"
+
+namespace datadog {
+namespace tracing {
+
+// Converts a boolean value to a string
+// The resulted string is either `true` or `false`
```
dgoffredo, line 12: I'd say "true" or "false", but the meaning is clear either way.

src/datadog/string_util.h
```diff
@@ -0,0 +1,29 @@
...
+std::string to_string(bool b);
```
dgoffredo, line 13: Since you don't refer to the variable name in the contract, then it's safe to omit it, especially here where the interpretation of the parameter is obvious.

Alternatively, you could go full pedant and write it this way:
```c++
// Return a string representation of the specified boolean `value`.
// The result is "true" for `true` and "false" for `false`.
std::string to_string(bool value);
```

src/datadog/string_util.cpp
```diff
@@ -0,0 +1,81 @@
...
+      case PropagationStyle::B3:
+        return "b3";
+      case PropagationStyle::DATADOG:
+        return "datadog";
+      case PropagationStyle::W3C:
+        return "tracecontext";
+      case PropagationStyle::NONE:
+        return "none";
+    }
+    return "";  ///< unlikely
```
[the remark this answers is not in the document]
dgoffredo: Also, `inline` could be `extern`, but then you need a name for the size of the array :angry: (and possibly a name for the `struct`).

dgoffredo (commented):
I just realized that I have two tabs open with two different code reviews. Flushing comments...

dgoffredo (commented):
Flushing comments...

dgoffredo (approved):
This will work, thanks for all of the thought put into it.

I'm wondering about alternative ways to factor `ConfigManager`'s data members. Right now, you have one data member for remotely configurable thing, and share some code via `class DynamicConfig`. Then the work of managing the new values is in `ConfigManager::update`.

You set `DD_TAGS` by reforming the `shared_ptr<const SpanDefaults>`, because those tags live in `SpanDefaults`.

What happens when we need to separately configure a different property of [SpanDefaults][1]? Well, we'll almost certainly never need to. The other properties are more about span identity than about span annotation.

Still, I wonder if another protocol is warranted.
```c++
// TODO: Bad name, 0/10, do not recommend.
class DynamicallyConfigurable {
public:
  virtual std::vector<ConfigMetadata> update(const ConfigUpdate&) = 0;
  virtual void reset() { update({}); } // shrug
};
```
The benefit is that you can organize things by having constructors take dependencies, if any.
```c++
class ConfigManager {
  shared_ptr<const SpanDefaults> span_defaults_;
  shared_ptr<TraceSampler> trace_sampler_;
  SampleRateManager sample_rate_manager_;
  SamplingRulesManager sampling_rules_manager_;
  TagsManager tags_manager_;
  FooManager foo_manager_;
  ReportTracesManager report_traces_manager_;

  // ...
public:

  ConfigManager(...)
  : sample_rate_manager_(&trace_sampler_)
  , sampling_rules_manager_(&trace_sampler_)
  , tags_manager_(&span_defaults_)
  , foo_manager_(&span_defaults_)
  { ... }
};

vector<ConfigMetadata> ConfigManager::update(const 
vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
  vector<ConfigMetadata> metadata;

  lock_guard lock(mutex_);
  sample_rate_manager_.update(conf);
  sampling_rules_manager_.update(conf);
  ...
```
The bookkeeping about the original value and such happens within each `DynamicallyConfigurable` derived class implementation.

I'm not confident that this is an improvement, but it's an idea.

Anyway, LGTM.

[1]: https://github.com/DataDog/dd-trace-cpp/blob/60eed64ebbf89e85276d31f99f222d880ef88ae4/src/datadog/span_defaults.h#L15

src/datadog/config_manager.h
```diff
@@ -10,23 +10,58 @@
...
 #include "json.hpp"
+#include "optional.h"
 #include "span_defaults.h"
 #include "tracer_config.h"
 
 namespace datadog {
 namespace tracing {
 
 class ConfigManager {
+  // A template class for managing dynamic configuration values.
```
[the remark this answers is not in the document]
dmehala: TIL

PR #99, opened 2024-02-28
SET(CURL_ZLIB OFF)

dgoffredo:
More recent Alpine images were failing to build nginx-datadog due to libz.so being in a different location on the file system than expected by the build produced by CMake.

We do not want to build (libcurl) with zlib anyway. Looking around, I found that CURL_ZLIB was needed in addition to (or instead of?) USE_ZLIB.


dmehala (approved):
Thank you! :shipit:

PR #102, opened 2024-03-13
chore(bazel): add MODULE.bazel files for bzlmod

mmorel-35:
Bzlmod is the new way to handle external dependencies.
This applies https://bazel.build/external/migration

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /bazel-*
 /.build/
 /.coverage/
+MODULE.bazel.lock
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,20 @@
+module(
+    name = "dd-trace-cpp",
+    version = "",
+)
+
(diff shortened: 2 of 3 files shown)


dgoffredo:
Thanks for the patch!

Our bazel build mode is for consumption by [Envoy][1]. Do you expect that their use of our code will continue to work after this change? I have yet to try it.

[1]: https://github.com/envoyproxy/envoy

mmorel-35:
> Do you expect that their use of our code will continue to work after this change? 

It is what claims Bazel, that bzlmod and workspace mode can work at the same time.
What I'm trying to do here is to make envoy relsolve its dependencies as mush as possible throught bzlmod.

dgoffredo:
Looks good to me. The coverage publishing failure is either spurious or related to a recent change to our CI.

PR #105, opened 2024-03-25
fix: split DD_TAGS either by space or comma

### LINKED ISSUE #88: `DD_TAGS` space separation
# Description

During my attempt to use `nginx-datadog` instead of the legacy nginx-opentracing module in system-tests. I found out `dd-trace-cpp` was crashing because of [this odd DD_TAGS expression](https://github.com/DataDog/system-tests/blob/47cdfce7fd9911f2828870fd69ae13db45da87fb/utils/build/docker/set-system-tests-weblog-env.Dockerfile#L6).

Let me save you a click (the link is here for context):
````dockerfile
ENV DD_TAGS='key1:val1, key2 : val2 '
````

## Investigation
I found this [RFC from 2021](https://docs.google.com/document/d/1LKOhM59ryAgdJNRKPgy-Id-wgCMpdMaJMMD9aLzQETU) to support space separation in `DD_TAGS`.

Our implementation does not follow the recommend solution of this RFC and end up mixing spaces and commas. The result of our algorithm is:
Screenshot 2023-12-21 at 17 09 22

While it should be as follow according to the RFC:
````cpp
const std::vector<StringView> tags {
   {"key1:val1"},
   {"key2 : val2"}
};
````

### PR DESCRIPTION
**Pull request by dmehala**
# Description

Resolves #88

### DIFF
--- a/src/datadog/parse_util.cpp
+++ b/src/datadog/parse_util.cpp
@@ -151,7 +151,6 @@ std::vector<StringView> parse_list(StringView input) {
   return items;
 }
 
-// TODO: normalize key and value
 Expected<std::unordered_map<std::string, std::string>> parse_tags(
     std::vector<StringView> list) {
   std::unordered_map<std::string, std::string> tags;
--- a/src/datadog/parse_util.cpp
+++ b/src/datadog/parse_util.cpp
@@ -9,6 +9,7 @@
 #include <string>
 
 #include "error.h"
+#include "string_util.h"
 
 namespace datadog {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/datadog/parse_util.cpp [hunk 1]
```diff
@@ -150,32 +151,87 @@ std::vector<StringView> parse_list(StringView input) {
   return items;
 }
 
+// TODO: normalize key and value
```
dmehala, line 154:
```suggestion
```

PR #106, opened 2024-03-28
feat: introduce the option to disable Remote Configuration

dmehala:
# Description
Disable Remote Configuration programmatically or using `DD_REMOTE_CONFIGURATION_ENABLED=0`.

Content:
- Add support for `DD_REMOTE_CONFIGURATION_ENABLED` environment variable
- Cease periodic request for Remote Configuration when the feature is disabled.
- Store and cancel scheduled tasks all at once.
- chore: Relocate string utility functions from `parse_util` to `string_util`.


src/datadog/datadog_agent_config.h
```diff
@@ -56,6 +56,8 @@ struct DatadogAgentConfig {
   Optional<int> request_timeout_milliseconds;
   // Maximum amount of time the process is allowed to wait before shutting down.
   Optional<int> shutdown_timeout_milliseconds;
+  // TBD
```
dmehala, line 59:
```suggestion
  // Enable the capability that allows to remotely configure and change the behavior of the tracer
```

PR #110, opened 2024-04-06
use abseil-cpp bazel_dep instead of extension

mmorel-35:


MODULE.bazel
```diff
@@ -3,18 +3,17 @@ module(
...
 
+bazel_dep(
+    name = "abseil-cpp",
+    version = "20240116.0",
+    repo_name = "com_google_absl",
+)
 bazel_dep(
     name = "bazel_skylib",
-    version = "1.2.1",
+    version = "1.5.0",
```
dmehala, line 13: Any specific reason to upgrade?

MODULE.bazel
```diff
@@ -3,18 +3,17 @@ module(
     version = "",
 )
 
+bazel_dep(
+    name = "abseil-cpp",
+    version = "20240116.0",
+    repo_name = "com_google_absl",
+)
```
dmehala, line 10: The extension was applying a patch. Any reason to change the behaviour? Is the patch part of `20240116.0`?

PR #111, opened 2024-04-10
Allow setting internal tags and expose numeric tags as metrics

Anilm3:
Simple PR which introduces the following changes:
- Expose `numeric_tags` through the API, although the new API functions refer to `metrics` for consistency with other tracers (e.g. PHP).
- Allow setting internal tags through the API, this is necessary as `dd-trace-cpp` is used by other internal projects such as ASM.

### DIFF
--- a/src/datadog/span_data.cpp
+++ b/src/datadog/span_data.cpp
@@ -51,9 +51,7 @@ void SpanData::apply_config(const SpanDefaults& defaults,
     tags.insert_or_assign(tags::version, version);
   }
   for (const auto& [key, value] : config.tags) {
-    if (!tags::is_internal(key)) {
-      tags.insert_or_assign(key, value);
-    }
+    tags.insert_or_assign(key, value);
   }
 
   resource = config.resource.value_or(name);
--- a/src/datadog/span.cpp
+++ b/src/datadog/span.cpp
@@ -124,9 +124,7 @@ void Span::set_metric(StringView name, double value) {
   data_->numeric_tags.insert_or_assign(std::string(name), value);
 }
 
-void Span::remove_tag(StringView name) {
-  data_->tags.erase(std::string(name));
-}
+void Span::remove_tag(StringView name) { data_->tags.erase(std::string(name)); }
 
 void Span::remove_metric(StringView name) {
   data_->numeric_tags.erase(std::string(name));
--- a/src/datadog/span.cpp
+++ b/src/datadog/span.cpp
@@ -101,10 +101,6 @@ const std::string& Span::name() const { return data_->name; }
 const std::string& Span::resource_name() const { return data_->resource; }
 
 Optional<StringView> Span::lookup_tag(StringView name) const {
-  if (tags::is_internal(name)) {
-    return nullopt;
-  }
-
   const auto found = data_->tags.find(std::string(name));
   if (found == data_->tags.end()) {
     return nullopt;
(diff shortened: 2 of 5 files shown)


dmehala:
I don't know much about metrics. Until then, is there a way to test/view metrics sent to the Agent in the Datadog UI? In other words, how did you test it and how can I?

dgoffredo:
"metrics" is a word used to describe floating point valued (`double`) span tags. Hence my choice of the name "numeric tag" instead of "metric."

They're sent as "metrics" (along with normal tags, called "meta") in the messagepack/protobuf message sent to the Agent and then to Intake:  https://github.com/DataDog/datadog-agent/blob/74fae3f4afeb77720e6584cccbbe853b5c9d677d/pkg/proto/datadog/trace/span.proto#L53-L55

_It has nothing to do with metrics. "Metrics" is a terrible name._ You can imagine how, historically, someone might have imagined it would be used for metrics, but it is not.

It is an open question which tags belong in "meta" (strings) and which tags belong in "metrics." There is some disagreement, e.g. for `http.status_code`.

My opinion is that tags are strings, and "metrics" is a ill-conceived attempt at optimization, or an artifact of some constraint imposed by some library in the Agent or Intake.

There's my unsolicited opinion. :)

src/datadog/span_data.cpp
```diff
@@ -51,9 +51,7 @@ void SpanData::apply_config(const SpanDefaults& defaults,
     tags.insert_or_assign(tags::version, version);
   }
   for (const auto& [key, value] : config.tags) {
-    if (!tags::is_internal(key)) {
-      tags.insert_or_assign(key, value);
-    }
```
dgoffredo, line 56: Do you want any user of the library to be able to set internal tags? Same question in `span.cpp`.

If Datadog has an internal client who wishes to have special access to the span representation, then consider adding an API specific to that use case, and document it as such.

Anilm3:
> I don't know much about metrics. Until then, is there a way to test/view metrics sent to the Agent in the Datadog UI? In other words, how did you test it and how can I?

This is what I used to test it:

```cpp
#include <datadog/span_config.h>
#include <datadog/tracer.h>
#include <datadog/tracer_config.h>

int main() {
    datadog::tracing::TracerConfig config;
    config.service = "test-service";
    config.environment = "staging";

    const auto validated_config = datadog::tracing::finalize_config(config);
    if (!validated_config) {
        return 1;
    }

    datadog::tracing::Tracer tracer{*validated_config};
    datadog::tracing::SpanConfig options;

    options.name = "metric-span";
    datadog::tracing::Span parent = tracer.create_span(options);

    parent.set_metric("metric", 10.0);

    return 0;
}
```

You'll need to have the agent running with the relevant API key, but you should be able to see a span containing the metric:

Screenshot from 2024-04-12 09-26-57

src/datadog/span_data.cpp
Anilm3: I'm not sure there is much value in forbidding internal tags, however I'm open to the idea of separating the `*_tag` and `*_metric` methods into `*_internal_tag` and `*internal_metric` if that makes it more palatable.

dmehala (approved):

dmehala (commented):
I do agree "metrics" are essentially just tags. I also think `*_numeric_tags` is better suited but I also understand consistency across tracers is key.  Thanks @dgoffredo to chiming in and sharing your knowledge 🙂

PR #112, opened 2024-04-13
fix: glob matching case sensitivity

dmehala:
# Description
Ensure glob matching is case insensitive and add test covering case sensitivity.

### DIFF
--- a/src/datadog/glob.cpp
+++ b/src/datadog/glob.cpp
@@ -1,5 +1,6 @@
 #include "glob.h"
 
+#include <cctype>
 #include <cstdint>
 
 namespace datadog {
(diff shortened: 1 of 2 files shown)


src/datadog/glob.cpp
```diff
@@ -18,8 +19,11 @@ bool glob_match(StringView pattern, StringView subject) {
   Index next_p = 0;  // next [p]attern index
   Index next_s = 0;  // next [s]ubject index
 
-  while (p < pattern.size() || s < subject.size()) {
-    if (p < pattern.size()) {
+  const size_t p_size = pattern.size();
```
Anilm3, line 22: nit: I can't comment on the relevant line, but instead of `using Index = std::size_t;` it would make more sense to use `using Index = StringView::size_type;`, it should work as expected with both `absl::string_view` and `std::string_view`. In practice it doesn't make a difference.

Anilm3 (approved):
lgtm

PR #114, opened 2024-04-22
parse_util: include cstdint to fix compilation under GCC 13

malt3:
Since GCC 13, std::uint64_t and friends require an explicit include of `<cstdint>`.
See also https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes


dmehala (approved):
LGTM

PR #115, opened 2024-04-29
feat: extract/inject datadog tracestate field `p`

dmehala:
# Description

Add support for W3C datadog tracestate field `p`.

### DIFF
--- a/examples/http-server/proxy/proxy.cpp
+++ b/examples/http-server/proxy/proxy.cpp
@@ -43,7 +43,7 @@ int main() {
 
   dd::Tracer tracer{*finalized_config};
 
-  httplib::Client upstream_client("server", 8080);
+  httplib::Client upstream_client("server", 80);
 
   // Configure the HTTP server.
   auto forward_handler = [&tracer, &upstream_client](
--- a/test/test_span.cpp
+++ b/test/test_span.cpp
@@ -795,7 +795,6 @@ TEST_CASE("injecting W3C tracestate header") {
   span->inject(writer);
 
   CAPTURE(writer.items);
-  // TODO: substitue parent ID in tracestate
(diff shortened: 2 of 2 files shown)


examples/http-server/proxy/proxy.cpp
```diff
@@ -43,43 +43,44 @@ int main() {
 
   dd::Tracer tracer{*finalized_config};
 
-  httplib::Client upstream_client("server", 80);
+  httplib::Client upstream_client("server", 8080);
```
dmehala, line 46:
```suggestion
  httplib::Client upstream_client("server", 80);
```

src/datadog/w3c_propagation.cpp
```diff
@@ -300,6 +308,7 @@ Expected<ExtractedData> extract_w3c(
     return result;
   }
 
+  result.datadog_w3c_parent_id = "0000000000000000";
```
pablomartinezbernardo, line 311: Curious to know: Is there a reason to use 0s id instead of an empty field somehow? Is the all 0s value a special value somewhere else?

pablomartinezbernardo (approved):
LGTM

src/datadog/w3c_propagation.cpp
dmehala: It's from the RFC. I think it's an optimization for the backend.

test/test_span.cpp
```diff
@@ -794,9 +795,13 @@ TEST_CASE("injecting W3C tracestate header") {
   span->inject(writer);
 
   CAPTURE(writer.items);
+  // TODO: substitue parent ID in tracestate
```
dmehala, line 798:
```suggestion
```

PR #116, opened 2024-04-29
feat: support Remote Config sampling rules

dmehala:


src/datadog/config_manager.cpp
```diff
@@ -35,32 +98,46 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
...
-    // This reset rate limiting and `TraceSampler` has no `operator==`.
-    // TODO: Instead of creating another `TraceSampler`, we should
-    // update the default sampling rate.
-    trace_sampler_ = std::move(trace_sampler);
-    metadata.emplace_back(std::move(trace_sampling_metadata));
+    auto maybe_rules = parse_trace_sampling_rules(*conf.trace_sampling_rules);
+    if (auto error = maybe_rules.if_error()) {
+      trace_sampling_rules_metadata.error = std::move(*error);
+    } else {
+      rules.merge(*maybe_rules);
```
pablomartinezbernardo, line 134: I'm looking at [the docs for this](https://en.cppreference.com/w/cpp/container/unordered_map/merge) and if I'm understanding correctly, items already in `rules` will not be overriden by those in `maybe_rules`. Is that the intended behavior?

pablomartinezbernardo (approved):
One small comment, otherwise LGTM

dmehala:
@pablomartinezbernardo 

Since your review I added 3 commits:
- 9545d64206d236f121ae311856d7398a146e1c82
- be97f1986e25e4fde626dde64fa5d9d4c29b139f
- 23bb27cc653b4cb64da4b31b497cc7b6c1c4e941

I fixed the rule override issue you mentioned but also remote configuration system tests revealed part of the implementation I overlooked.

src/datadog/span_sampler_config.cpp
```diff
@@ -16,7 +16,12 @@ namespace {
 std::string to_string(const std::vector<SpanSamplerConfig::Rule> &rules) {
   nlohmann::json res;
   for (const auto &r : rules) {
-    res.emplace_back(r.to_json());
+    auto j = r.to_json();
+    j["sample_rate"] = r.sample_rate;
+    if (r.max_per_second) {
+      j["max_per_second"] = *r.max_per_second;
+    }
+    res.emplace_back(std::move(j));
```
pablomartinezbernardo, line 24: Curious to know, as they are being set in `to_json`, why is it necessary to set them again?

src/datadog/trace_sampler_config.cpp
```diff
@@ -196,6 +197,9 @@ Expected<FinalizedTraceSamplerConfig> finalize_config(
     result.metadata[ConfigName::TRACE_SAMPLING_RATE] = ConfigMetadata(
         ConfigName::TRACE_SAMPLING_RATE, to_string(*sample_rate, 1),
         ConfigMetadata::Origin::CODE);
+  } else {
+    result.metadata[ConfigName::TRACE_SAMPLING_RATE] = ConfigMetadata(
+        ConfigName::TRACE_SAMPLING_RATE, "1.0", ConfigMetadata::Origin::CODE);
```
pablomartinezbernardo, line 202: Wanted `ConfigMetadata::Origin::DEFAULT` here? Not sure if it has any side effect, and it's true that it is `CODE` as well though

pablomartinezbernardo (approved):

src/datadog/span_sampler_config.cpp
dmehala: `to_json` is not overloaded by `SpanSamplerConfig::Rule`, meaning `SpanMatcher::to_json` is called instead.

src/datadog/trace_sampler_config.cpp
dmehala: Good catch. I did it right but I must have lost that part during a rebase :/

dmehala:
```suggestion
        ConfigName::TRACE_SAMPLING_RATE, "1.0", ConfigMetadata::Origin::DEFAULT);
```

PR #118, opened 2024-04-30
feat: Report host information in telemetry

dmehala:
# Description
The telemetry implementation was missing some host information. This PR adds those.

Content:
- Collect informations on the Operating System and the CPU architecture
- Report those informations in telemetry

### DIFF
--- a/src/datadog/datadog_agent.cpp
+++ b/src/datadog/datadog_agent.cpp
@@ -18,7 +18,6 @@
 #include "string_view.h"
 #include "trace_sampler.h"
 #include "tracer.h"
-#include "version.h"
 
 namespace datadog {
 namespace tracing {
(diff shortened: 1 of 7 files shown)


src/datadog/datadog_agent.cpp
```diff
@@ -366,9 +368,27 @@ void DatadogAgent::flush() {
...
+void DatadogAgent::send_telemetry(StringView request_type,
+                                  std::string payload) {
+  auto set_telemetry_headers = [request_type, &payload,
+                                debug_enabled = tracer_telemetry_->debug(),
+                                tracer_signature =
+                                    &tracer_signature_](DictWriter& headers) {
+    /*
+      TODO:
+        Datadog-Container-ID
+    */
```
dmehala, line 380: Todo or not todo? That's the question. 🤔

src/datadog/tracer_telemetry.h
```diff
@@ -48,6 +49,7 @@ class TracerTelemetry {
   bool debug_ = false;
   Clock clock_;
   std::shared_ptr<Logger> logger_;
+  HostInfo host_info_;
   TracerSignature tracer_signature_;
   std::string hostname_;
```
pablomartinezbernardo, line 54: We should be able to remove `hostname_`, right?

pablomartinezbernardo (approved):
Small comment, LGTM

src/datadog/datadog_agent.cpp
dmehala: Won't do

PR #119, opened 2024-05-06
feat: add windows support

dmehala:
# Description
This PR add support for windows. It is a prerequisite for #85.

Contains: 
- refactor: the codebase to use `substr` instead of `range`.
- refactor: CMake targets.
- refactor: add bazelrc to build using either abseil or std::string STD.
- ci: build and run on windows using CMake and bazel.
- update: add platform support section and update building instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,12 @@ int main() {
 See the [examples](examples) directory for more extensive usage examples.
 
 ## Platform Support
-The library is currently compatible with and has been tested on the following CPU architecture, OS and compiler combinations:
+The library has been tested and is compatible on the following CPU architecture, OS and compiler combinations:
 - x86_64 and arm64 Linux with GCC 11.4.
 - x86_64 and arm64 Linux with Clang 14.
 - x86_64 Windows with MSVC 2022.
 - arm64 macOS with Apple Clang 15.
 
-All possible configurations cannot be tested. However, any conforming C++17 compiler should be supported.
 
 ## Building and Installation
 
@@ -67,7 +66,7 @@ cmake --build build -j
 ```
 
 By default CMake will generate both static and shared libraries. To build either on of them use
-as configuration parameter `BUILD_SHARED_LIBS` or `BUILD_STATIC_LIBS`. Example:
+either `BUILD_SHARED_LIBS` or `BUILD_STATIC_LIBS`. Example:
 
 ```shell
 cmake -B build -DBUILD_SHARED_LIBS=1 .
--- a/README.md
+++ b/README.md
@@ -38,64 +38,81 @@ int main() {
 ```
 See the [examples](examples) directory for more extensive usage examples.
 
-Build
------
-### `cmake && make && make install` Style Build
-Build this library from source using [CMake][1]. Installation places a shared
-library and public headers into the appropriate system directories
-(`/usr/local/[...]`), or to a specified installation prefix.
+## Platform Support
+The library is currently compatible with and has been tested on the following CPU architecture, OS and compiler combinations:
+- x86_64 and arm64 Linux with GCC 11.4.
+- x86_64 and arm64 Linux with Clang 14.
+- x86_64 Windows with MSVC 2022.
+- arm64 macOS with Apple Clang 15.
 
-A recent version of CMake is required (3.24), which might not be in your
+All possible configurations cannot be tested. However, any conforming C++17 compiler should be supported.
+
+## Building and Installation
+
+### Requirements
+`dd-trace-cpp` requires a [supported](#platform-support) C++17 compiler.
+
+A recent version of CMake is required (`3.24`), which might not be in your
 system's package manager. [bin/install-cmake](bin/install-cmake) is an installer
 for a recent CMake.
 
-dd-trace-cpp requires a C++17 compiler.
+### Building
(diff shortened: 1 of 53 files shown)


README.md
```diff
@@ -38,64 +38,81 @@ int main() {
...
-library and public headers into the appropriate system directories
-(`/usr/local/[...]`), or to a specified installation prefix.
+## Platform Support
+The library is currently compatible with and has been tested on the following CPU architecture, OS and compiler combinations:
+- x86_64 and arm64 Linux with GCC 11.4.
+- x86_64 and arm64 Linux with Clang 14.
+- x86_64 Windows with MSVC 2022.
+- arm64 macOS with Apple Clang 15.
+
+All possible configurations cannot be tested. However, any conforming C++17 compiler should be supported.
```
dmehala, line 48:
```suggestion
All possible configurations cannot be tested. However, any conforming C++17 compiler should be supported.
```

README.md
```diff
@@ -38,64 +38,81 @@ int main() {
...
+
+```shell
+git clone 'https://github.com/datadog/dd-trace-cpp'
+cd dd-trace-cpp
+cmake -B build .
+cmake --build build -j
+```
+
+By default CMake will generate both static and shared libraries. To build either on of them use
+as configuration parameter `BUILD_SHARED_LIBS` or `BUILD_STATIC_LIBS`. Example:
```
dmehala, line 70:
```suggestion
either `BUILD_SHARED_LIBS` or `BUILD_STATIC_LIBS`. Example:
```

src/datadog/expected.h
```diff
@@ -99,15 +99,21 @@ class Expected {
...
+template <typename Value>
+Expected<Value>::Expected(const Error& error) : data_(error) {}
+
+template <typename Value>
+Expected<Value>::Expected(Error&& error) : data_(std::move(error)) {}
+
+// Expected<Value>& Expected<Value>::operator=(Other&& other) {
+//   data_ = std::forward<Other>(other);
+//   return *this;
+// }
```
dmehala, line 116:
```suggestion

// Expected<Value>& Expected<Value>::operator=(Other&& other) {
//   data_ = std::forward<Other>(other);
//   return *this;
// }
```

test/system-tests/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+# add_library(httplib INTERFACE)
+# target_include_directories(httplib SYSTEM INTERFACE .)
+
```
dmehala, line 3:
```suggestion
```

test/system-tests/main.cpp
```diff
@@ -160,7 +160,7 @@ int main(int argc, char* argv[]) {
     res.status = 500;
   });
 
-  std::signal(SIGTERM, hard_stop);
+  // std::signal(SIGTERM, hard_stop);
```
dmehala, line 163: - [ ] TODO

pablomartinezbernardo (approved):
Small comment, LGTM

src/datadog/string_util.cpp
```diff
@@ -79,12 +88,15 @@ std::string join_tags(
...
+  const auto prefix_end = p + prefix.size();
+  while (*s == *p) {
+    ++s;
+    ++p;
   }
 
-  return std::mismatch(subject.begin(), subject.end(), prefix.begin()).second ==
-         prefix.end();
+  return p == prefix_end;
 }
```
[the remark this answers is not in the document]
dmehala: [question] why?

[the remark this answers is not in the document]
pablomartinezbernardo: There's nothing stopping `s` and `p` from going over the end of the `string_view`, so it would be undefined behavior, but even if nothing breaks the condition `p == prefix_end` will be false. Did a small example to reproduce it

image


To avoid that, the function could be written like this for example

```cpp
bool starts_with(StringView subject, StringView prefix) {
  auto s = subject.data();
  auto p = prefix.data();
  const auto prefix_end = p + prefix.size();
  while (p < prefix_end) {
    if (*s++ != *p++) {
        return false;
    }
  }

  return true;
}
```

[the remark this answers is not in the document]
dmehala: > There's nothing stopping s and p from going over the end of the string_view, so it would be undefined behavior
Indeed, thanks to Github diff awful diff I thought `if (prefix.size() > subject.size()) { ... }`  was part of the implementation, which reduce the area of the UB. However, as you mentioned, if `prefix.size() == subject.size()` then it's undefined behaviour and plain wrong. Thank you.

src/datadog/platform_util.cpp
```diff
@@ -91,7 +91,86 @@ HostInfo _get_host_info() {
...
+  RtlGetVersion func =
+      (RtlGetVersion)GetProcAddress(GetModuleHandleA("ntdll"), "RtlGetVersion");
+
+  if (func) {
+    OSVERSIONINFOEXW os_info;
+    ZeroMemory(&os_info, sizeof(OSVERSIONINFO));
+    os_info.dwOSVersionInfoSize = sizeof(os_info);
+
+    if (func(&os_info) == 0) {
+      // major, minor, build -> name, version
```
dmehala, line 112:
```suggestion
```

PR #123, opened 2024-05-22
fix: `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS` support floating point input

dmehala:
# Description

According to the Remote Configuration specification, `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS` support floating point input and can be set to `0`.

# Motivation

System-tests use sub-seconds value for fast iteration. In order to test the library against remote configuration tests, the library must support sub-seconds inputs.

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/src/datadog/datadog_agent_config.cpp
+++ b/src/datadog/datadog_agent_config.cpp
@@ -21,7 +21,7 @@ Expected<DatadogAgentConfig> load_datadog_agent_env_config() {
 
   if (auto raw_rc_poll_interval_value =
           lookup(environment::DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS)) {
-    auto res = parse_int(*raw_rc_poll_interval_value, 10);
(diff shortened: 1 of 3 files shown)


src/datadog/datadog_agent_config.cpp
```diff
@@ -21,7 +21,7 @@ Expected<DatadogAgentConfig> load_datadog_agent_env_config() {
 
   if (auto raw_rc_poll_interval_value =
           lookup(environment::DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS)) {
-    auto res = parse_int(*raw_rc_poll_interval_value, 10);
+    auto res = parse_double(*raw_rc_poll_interval_value);
```
dmehala, line 24: Note to myself: a double is too big but I don't have time (don't want yet) to write (and test) a `parse_float` function.

pablomartinezbernardo (approved):

PR #125, opened 2024-05-27
fix: trace sampling rules order

dmehala:
# Description

Reset trace sampling rules legacy order.

### DIFF
--- a/src/datadog/config_manager.cpp
+++ b/src/datadog/config_manager.cpp
@@ -57,7 +57,7 @@ Expected<Rules> parse_trace_sampling_rules(const nlohmann::json& json_rules) {
                      std::move(message)};
       }
 
-      auto provenance = to_lower(provenance_it->get<std::string_view>());
+      auto provenance = to_lower(provenance_it->get<StringView>());
       if (provenance == "customer") {
         rule.mechanism = SamplingMechanism::REMOTE_RULE;
       } else if (provenance == "dynamic") {
--- a/src/datadog/config_manager.cpp
+++ b/src/datadog/config_manager.cpp
@@ -110,8 +110,8 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
 
   // NOTE(@dmehala): Sampling rules are generally not well specified.
   //
-  // Rules are eveluated in the order they are inserted, which means the most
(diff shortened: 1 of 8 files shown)


src/datadog/config_manager.cpp
```diff
@@ -98,7 +108,19 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
 
   std::lock_guard<std::mutex> lock(mutex_);
 
-  decltype(rules_) rules;
+  // NOTE(@dmehala): Sampling rules are generally not well specified.
+  //
+  // Rules are eveluated in the order they are inserted, which means the most
```
dmehala, line 113:
```suggestion
  // Rules are evaluated in the order they are inserted, which means the most
```

src/datadog/config_manager.cpp
```diff
@@ -42,7 +43,10 @@ Expected<Rules> parse_trace_sampling_rules(const nlohmann::json& json_rules) {
         return *error;
       }
 
-      rate.value = *maybe_rate;
+      rule.rate = *maybe_rate;
+    } else {
+      return Error{Error::TRACE_SAMPLING_RULES_INVALID_JSON,
+                   "Missing \"sample_rate\" field"};
```
pablomartinezbernardo, line 49: qq for this and other new error handling: Is a default value set somewhere? Or will parsing now fail when it didn't before?

src/datadog/config_manager.cpp
```diff
@@ -98,7 +108,19 @@ std::vector<ConfigMetadata> ConfigManager::update(const ConfigUpdate& conf) {
 
   std::lock_guard<std::mutex> lock(mutex_);
 
-  decltype(rules_) rules;
+  // NOTE(@dmehala): Sampling rules are generally not well specified.
+  //
+  // Rules are eveluated in the order they are inserted, which means the most
+  // specific matching rule moght not be evaluated, even though it should be.
```
pablomartinezbernardo, line 114:
```suggestion
  // specific matching rule might not be evaluated, even though it should be.
```

pablomartinezbernardo (approved):
LGTM

src/datadog/config_manager.cpp
dmehala: Parsing will fail, those are mandatory fields.

PR #126, opened 2024-05-27
chore: publish coverage to codecov

dmehala:
# Description

Publish code coverage to codecov.

# Motivation

To standardize with other datadog repos and it's easier to use.


bin/test
```diff
@@ -58,16 +58,17 @@ if [ "$coverage_flags" != '' ]; then
     mkdir -p .coverage
     # Compile the .gcdo files into an intermediate format.
     # Use the faster JSON Perl module, JSON:XS, to speed things up.
-    lcov --quiet --capture --directory .build --output-file .coverage/raw.info --rc geninfo_json_module=JSON::XS
+    lcov --quiet --capture --directory .build --output-file .coverage/raw.info #--rc geninfo_json_module=JSON::XS
```
dmehala, line 61:
```suggestion
    lcov --quiet --capture --directory .build --output-file .coverage/raw.info --rc geninfo_json_module=JSON::XS
```

pablomartinezbernardo: This is probably exactly why you commented it: to try it out. Did you find where `geninfo_json_module` is defined? I couldn't find any info

pablomartinezbernardo (approved):

bin/test
dmehala: It is working but codecode would not consume `.coverage/raw.info`, I thought maybe codecov is struggling to parse the JSON output.

PR #130, opened 2024-06-14
refactor: Remote Configuration

dmehala:
## Description
Refactor the Remote Configuration (RC) as a standalone module, enabling its use independently of the tracer. Additionally, ensure that external listeners can register and receive configuration updates when RC is embedded in the tracer.

## Motivation
ASM needs a way to receives RC updates for NGINX integration.

## Changes:
  - Move APM_TRACING updates handling to the configuration manager.
  - Expose the `remote_configuration_listeners` field in datadog agent config to allow registration of external listeners.
  - Improve the robustness of config path parsing and ensure an error is reported if parsing fails.
  - Add `to_upper` utility function.
  - Increase test coverage for RC and the new configuration manager behaviour.


src/datadog/remote_config/capability.h
```diff
@@ -0,0 +1,130 @@
+#pragma once
```
dmehala, line 1: Split the file: `product.h` and `capability.h`

src/datadog/remote_config/capability.h
```diff
@@ -0,0 +1,130 @@
...
+//
+// using namespace datadog::remote_config::product;
+// Products p = AGENT_CONFIG | APM_TRACING;
+using Products = uint64_t;
+
+namespace capability {
+
+enum Flag : Capabilities {
+  // DEFAULT is a special value. It is the default of the product capability.
+  DEFAULT = 0,
```
dmehala, line 32: Remove `DEFAULT`.

pablomartinezbernardo (approved):

src/datadog/remote_config/remote_config.cpp
```diff
@@ -230,48 +235,46 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
...
       }
     }
+
+    for (const auto& listener : listeners_) {
+      listener->on_post_process();
+    }
   } catch (const nlohmann::json::exception& e) {
-    std::string error_message = "Ill-formatted Remote Configuration response: ";
+    std::string error_message = "Invalid Remote Configuration response: ";
     error_message += e.what();
```
dmehala, line 274: log the error

src/datadog/remote_config/capability.h
```diff
@@ -0,0 +1,130 @@
...
+  X(LIVE_DEBUGGING, 4)                 \
+  X(LIVE_DEBUGGING_SYMBOL_DB, 5)       \
+  X(ASM, 6)                            \
+  X(ASM_DD, 7)                         \
+  X(ASM_DATA, 8)                       \
+  X(ASM_FEATURES, 9)
+
+namespace product {
+
+enum Flag : Products {
```
cataphract, line 83: a flag is by definition a boolean, so the naming is semantically off. I get that they're assigned a specific bit, but that only makes an implementation of a set of products easier, it doesn't make them logically a "flag"

src/datadog/remote_config/remote_config.cpp
```diff
@@ -173,28 +159,41 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
     if (client_configs_it == json.cend()) {
       if (!applied_config_.empty()) {
         std::for_each(applied_config_.cbegin(), applied_config_.cend(),
```
cataphract, line 161: configurations should be applied by config key alphabetical order, not by the order them they appear in the response

src/datadog/remote_config/remote_config.cpp
```diff
@@ -173,28 +159,41 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
...
+                        }
                       });
         applied_config_.clear();
       }
-      return config_updates;
+
+      for (const auto& listener : listeners_) {
+        listener->on_post_process();
+      }
+      return;
```
cataphract, line 174: configurations should be reverted when they are not included in a response, not when **all** configurations are removed from a response. It is possible to simultaneously apply a new configuration and revert a previous configuration.

src/datadog/remote_config/remote_config.cpp
```diff
@@ -173,28 +159,41 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
...
+                             listeners_per_product_[it.second.product]) {
+                          listener->on_revert(it.second);
+                        }
                       });
         applied_config_.clear();
       }
-      return config_updates;
+
+      for (const auto& listener : listeners_) {
+        listener->on_post_process();
```
cataphract, line 172: idem (1st part of the previous comment)

src/datadog/remote_config/remote_config.cpp
```diff
@@ -173,28 +159,41 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
...
       if (!applied_config_.empty()) {
         std::for_each(applied_config_.cbegin(), applied_config_.cend(),
-                      [this, &config_updates](const auto it) {
-                        auto updated = revert_config(it.second);
-                        config_updates.insert(config_updates.end(),
-                                              updated.begin(), updated.end());
+                      [this](const auto& it) {
+                        for (const auto& listener :
+                             listeners_per_product_[it.second.product]) {
+                          listener->on_revert(it.second);
```
cataphract, line 165: `on_revert` is not marked `noexception`, yet here its exceptions can bubble freely. not that rcte2 doesn't provide a way to report exceptions when an error occurs while reverting, so arguably they should be reported globally.

src/datadog/remote_config/remote_config.cpp
```diff
@@ -230,48 +235,46 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
...
         new_config.error_message = "Wrong service targeted";
       } else {
-        new_config.state = Configuration::State::acknowledged;
-        new_config.content = parse_dynamic_config(config_json.at("lib_config"));
-
-        auto updated = apply_config(new_config);
-        config_updates.insert(config_updates.end(), updated.begin(),
-                              updated.end());
+        for (const auto& listener : listeners_per_product_[*product]) {
+          // Q: Two listeners on the same product. What should be the behaviour
```
cataphract, line 239: this is only a problem if the errors occur on the same config id, not on the same product

src/datadog/remote_config/remote_config.cpp
```diff
@@ -45,68 +29,74 @@ constexpr std::array<uint8_t, sizeof(uint64_t)> capabilities_byte_array(
...
 }
 
-nlohmann::json RemoteConfigurationManager::make_request_payload() {
+nlohmann::json Manager::make_request_payload() {
   // clang-format off
   auto j = nlohmann::json{
     {"client", {
       {"id", client_id_},
-      {"products", nlohmann::json::array({k_apm_product})},
+      {"products", products_},
```
cataphract, line 97: You don't seem to be using `cached_target_files`, which results in the same configuration being sent over and over again.

cataphract (changes requested):

src/datadog/remote_config/listener.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual void on_revert(const Configuration&) = 0;
+
+  // Pure virtual function called when a configuration is updated.
+  // Returns an error message if the configuration could not be applied, or
+  // nothing.
+  virtual tracing::Optional<std::string> on_update(const Configuration&) = 0;
+
+  // TODO: Find a better name and a better solution! Mainly here for ASM
+  // Called once the last remote configuration response is completed.
+  virtual void on_post_process() = 0;
```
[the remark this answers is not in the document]
dmehala: This is achievable by registering one listener for all the products needed, then `on_post_process()` will be called for that listener.

src/datadog/remote_config/remote_config.cpp
dmehala: I did not found any mention of the iteration order, can you indicate where is it specified?

src/datadog/config_manager.h
```diff
@@ -8,16 +8,31 @@
...
 #include "tracer_config.h"
+#include "tracer_telemetry.h"
 
 namespace datadog {
 namespace tracing {
 
-class ConfigManager {
+class ConfigManager : public remote_config::Listener {
+ public:
+  // The `ConfigUpdate` struct serves as a container for configuration that can
```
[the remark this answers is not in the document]
dmehala:
```suggestion
  // The `Update` struct serves as a container for configuration that can
```

src/datadog/remote_config/remote_config.cpp
dmehala: I deferred this until I am able to get an answer from RC people. My observation is configurations are not sent over and over again.

src/datadog/remote_config/remote_config.cpp
dmehala: This part of the code is executed when `client_configs` is absent from the response. I can find on the spec:
> client_configs: List of configurations path that the tracer MUST apply. 

My understand is configurations outside of `client_configs` are not intended for the tracer. Since the request is made with anything necessary to identify a tracer (`is_tracer` and `client_tracer`), I would expect the response to target the tracer. Am I missing something?

src/datadog/remote_config/capability.h
dmehala: Maybe, I am not fluent enough to argue. However, it is not uncommon to name bitwise enums `somethingFlags`. It is even a concept in C# ([doc](https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-8.0), [example](https://code-corner.dev/2023/11/04/Understanding-Flag-Enums-in-C/))

src/datadog/remote_config/remote_config.cpp
Anilm3: I think the RFC is quite clear here:

> Used to avoid sending the same config files over and over on the wire. Note that the agent does not make any effort trying to track this for the tracer and will resend every file in every update if this array is empty.

I don't think observations are relevant, the remote configuration client must meet the specification regardless of the idiosyncrasies of the agent.

src/datadog/remote_config/capability.h
```diff
@@ -0,0 +1,130 @@
...
+
+inline tracing::StringView to_string_view(product::Flag product) {
+#define X(NAME, ID)           \
+  case product::Flag::NAME: { \
+    return DD_QUOTED(NAME);   \
+  }
+  switch (product) { DD_LIST_REMOTE_CONFIG_PRODUCTS }
+#undef X
+
+  std::abort();
```
Anilm3, line 98: I know this is quite unlikely but abort is extreme as it can't be recovered from, perhaps an exception would be better?

dmehala: If it does happen I don't think an exception can help much. The program is in a state where the continuity and integrity of it is questionable.

src/datadog/remote_config/remote_config.cpp
```diff
@@ -173,28 +159,41 @@ std::vector<ConfigMetadata> RemoteConfigurationManager::process_response(
...
-      return config_updates;
+
+      for (const auto& listener : listeners_) {
+        listener->on_post_process();
+      }
+      return;
     }
 
     // Keep track of config path received to know which ones to revert.
     std::unordered_set<std::string> visited_config;
```
Anilm3, line 178: I'm not familiar with this json library, but if the string views you're using down below persist after the loop it might be worth changing this as well to use views.

src/datadog/remote_config/remote_config.cpp
Anilm3: This seems correct to me, I see that the code below handles the case @cataphract is describing...

src/datadog/remote_config/remote_config.cpp
dmehala: Good catch. Thank you.

src/datadog/remote_config/capability.h
Anilm3: Certainly that applies to the remote config client, but otherwise you can't make that assertion beyond that.

src/datadog/config_manager.cpp
```diff
@@ -77,16 +77,71 @@ Expected<Rules> parse_trace_sampling_rules(const nlohmann::json& json_rules) {
...
+      report_traces_(config.report_traces),
+      telemetry_(telemetry) {}
+
+rc::Products ConfigManager::get_products() { return rc::product::APM_TRACING; }
+
+rc::Capabilities ConfigManager::get_capabilities() {
+  using namespace rc::capability;
+  return APM_TRACING_SAMPLE_RATE | APM_TRACING_TAGS | APM_TRACING_ENABLED |
+         APM_TRACING_SAMPLE_RULES;
+};
```
Anilm3, line 128:
```suggestion
}
```
Stray `;` is causing the benchmarks to fail.

src/datadog/remote_config/remote_config.cpp
cataphract: I can't seem to find it, there are now more than 1000 documents referencing remote config... I didn't dream it, but maybe it applies only to the order for merging configuration (data, rules, etc) for AppSec. Since appsec applies config only at the end, and it merges the data only then, the iteration order _here_ doesn't actually make much of a difference. If I find the reference, I'll leave it here.

src/datadog/remote_config/remote_config.cpp
cataphract: Ah yes, I see there are two paths that handle `on_revert`, and I hadn't read into the latter one later in the function. I would refactor this to merge the two branches together. Also, this function is on the long side, it could be refactored.

src/datadog/remote_config/capability.h
dmehala: resolved in a32272c1539e812f597285b763d8d4f5fe93df94

src/datadog/remote_config/capability.h
dmehala: resolved in 423024471029e6be02f281ea7e2f134658d7f63d

src/datadog/remote_config/remote_config.cpp
dmehala: resolved in d547482a2d932db2a50043836caf8353aa4ae12b

src/datadog/config_manager.h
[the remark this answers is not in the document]
dmehala: resolved in d547482a2d932db2a50043836caf8353aa4ae12b

src/datadog/remote_config/remote_config.cpp
dmehala: `noexception` specifier is not a compile-time check guaranteeing a function is not throwing. It is the responsibility of the implementer to ensure no exceptions are thrown, otherwise, it will result in a call to `std::terminate`, which we cannot recover.  Here is an example:

```cpp
int do_stuff() {
   std::vector<int> v;
   return  v.at(1);
}

void f() noexcept {
   auto a = do_stuff();
}
```

All calls to `on_revert` are in a try-catch block, any exception from the user code will be catched. However I do agree I need to catch `std::exception` for that.

src/datadog/remote_config/remote_config.cpp
dmehala: addresed in d547482a2d932db2a50043836caf8353aa4ae12b

src/datadog/remote_config/capability.h
dmehala: I don't agree but I can understand how frightening `std::abort` can be. I addressed your concern in d547482a2d932db2a50043836caf8353aa4ae12b

src/datadog/remote_config/remote_config.cpp
dmehala: Addressed in a32272c1539e812f597285b763d8d4f5fe93df94

src/datadog/config_manager.cpp
dmehala: addressed in d547482a2d932db2a50043836caf8353aa4ae12b. Thank you.

src/datadog/remote_config/capability.h
cataphract: That is true enough, but the causation is the other way around, because it is a flag (a boolean), it tends to be a assigned a specific bit when making sets (bitsets) of these. If the product was used only in sets of products, to identify whether a certain product is present, then the usage would be acceptable. The [dictionary of computing](https://www.oxfordreference.com/display/10.1093/oi/authority.20110803095822224) defines a flag as:

> A variable whose value indicates the attainment of some designated state or condition by an item of equipment or a program. The flag is subsequently used as a basis for conditional branching and similar decision processes.

Say, in `extern _Atomic int interrupted; ...; while (!interrupted) { ... }`.

Anyway, at the end of the day it doesn't really matter, but I did find it confusing at first.

src/datadog/remote_config/remote_config.cpp
cataphract: I should have been clearer; what I meant was that I see a virtual function not marked `noexcept`, and unless told otherwise, I assume that its implementations can throw (if they can't, why wouldn't they be marked `noexcept`, thereby generating better machine code). Yet the calling code is not prepared for exceptions from the implementations.

dmehala: This has been addressed in edd5397dc1140322035c7f9af2f4f45e3584cf2c

cataphract (approved):

Anilm3 (approved):

PR #131, opened 2024-06-14
api: update the tracer api

dmehala:
## Description
Update the API:
- `extract_or_create_span` always returns a span, either the extracted one or a new span if the extraction fail.
- `override_sampling_priority` accepts `SamplingPriority`. 

## Motivation
`create_span` always returns a span, while `extract_span` may or may not return a span. Logically, one can think the behavior of `extract_or_create_span` should first attempt to `extract_span`, and if the extraction fails, it should then call `create_span`. At least, that's what the name of the function suggest and I think the desired behavior. 

The other one is for simplicity's sake.

### DIFF
--- a/src/datadog/tracer.h
+++ b/src/datadog/tracer.h
@@ -76,9 +76,9 @@ class Tracer {
   // new trace (see `create_span`).  Optionally specify a `config` indicating
   // the attributes of the span.  If a failure occurs, then return an error.
   // Note that the absence of a span to extract is not considered an error.
-  Expected<Span> extract_or_create_span(const DictReader& reader);
-  Expected<Span> extract_or_create_span(const DictReader& reader,
-                                        const SpanConfig& config);
+  Span extract_or_create_span(const DictReader& reader);
+  Span extract_or_create_span(const DictReader& reader,
+                              const SpanConfig& config);
 
   // Return a JSON object describing this Tracer's configuration. It is the same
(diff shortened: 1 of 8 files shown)


pablomartinezbernardo (approved):
What is our policy for incompatible changes in the sdk? We are not following semver right?

dmehala:
I think we used to follow semver. At least, that's the policy we are following rn.

src/datadog/tracer.h
```diff
@@ -76,9 +76,9 @@ class Tracer {
   // new trace (see `create_span`).  Optionally specify a `config` indicating
   // the attributes of the span.  If a failure occurs, then return an error.
   // Note that the absence of a span to extract is not considered an error.
```
dgoffredo, line 78:
```c++
  // Return a span extracted from the specified `reader` (see `extract_span`).
  // If there is no span to extract, or if an error occurs during extraction,
  // then return a span that is the root of a new trace (see `create_span`).
  // Optionally specify a `config` indicating the attributes of the span.
```
Fight the entropy!

> At least, that's what the name of the function suggest and I think the desired behavior

Desired by whom? The advantage of allowing extraction errors to be returned is that the caller can, if desired, log the error. I anticipated "traces/sampling broken in Istio thingy" bugs whose root cause is extraction failing without making any noise. Likely you weighed this possibility and decided the benefit of `Expected<Span> → Span` outweighed the risk.

PR #132, opened 2024-06-17
fix: prevent crash due to invalid headers

dmehala:
## Description

When the tracer encounters an invalid propagation header value, logging the list of headers and their values can lead to a crash.

## Details
For each key and value in the list of headers, a `nlohmann::json` instance is created with the following code:
```cpp
stream << nlohmann::json(it->first + ": " + it->second);
```

It is possible to craft an input for any of the headers inspected by the tracer that can cause a crash when the JSON library attempts to parse or dump the input.

### DIFF
--- a/src/datadog/extraction_util.cpp
+++ b/src/datadog/extraction_util.cpp
@@ -7,7 +7,6 @@
 #include <unordered_map>
 
 #include "extracted_data.h"
-#include "json.hpp"
 #include "logger.h"
 #include "parse_util.h"
 #include "string_util.h"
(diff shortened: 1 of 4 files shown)


henrymwang (approved):

bm1549 (commented):
Can we include some unit tests that fail for the old behavior, but pass for the new behavior?

dgoffredo:
> Can we include some unit tests that fail for the old behavior, but pass for the new behavior?

`nlohmann::json` might fail to encode a string? That's surprising.

Anyway, if the "crash" was due to an exception thrown by the JSON library, then [this][1] might help in testing.

[1]: https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#exceptions

pablomartinezbernardo (approved):

dmehala:
@dgoffredo Yes, it's a character encoding / serializing issue. 

> In the past I've used it as a shorthand for "give me a quoted, escaped string." 
I was trying to understand what was the purpose, thanks for the context. I will definitely audit for other uses of `nlohmann::json`

PR #136, opened 2024-06-26
chore: remove nlohmann::json from public headers

dmehala:
# Description

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


src/datadog/span_sampler.cpp
```diff
@@ -69,12 +69,13 @@ SpanSampler::Rule* SpanSampler::match(const SpanData& span) {
...
   std::vector<nlohmann::json> rules;
   for (const auto& rule : rules_) {
-    rules.push_back(to_json(rule));
+    rules.push_back(to_string(rule));
   }
 
   return nlohmann::json::object({
-      {"rules", rules},
-  });
+                                    {"rules", rules},
```
dgoffredo, line 76: Note that this change, and others made elsewhere, will cause the array element (in this case) to be quoted, since you're turning it into a string, e.g.

    [{"foo": 1}, {"bar": 2}]

will become

    ["{\"foo\":1}", "{\"bar\": 2}"]

This is one of the reasons (maybe the only reason) that I exposed the `nlohmann::json` type rather than using a `string` or inventing some third representation.

There were two reasons for the "pretty print the config as JSON" feature: that's how it was done in `dd-opentracing-cpp`, and that was an existing feature of `nginx-opentracing` that I preserved in `nginx-datadog`.

An alternative interface might be to invent a non-JSON human-only format for configuration, for example something based on indentation:

    void Foo::print_config(std::ostream& out, int indent_level = 0) {
        // ...
    }

In the end, I preferred JSON for its portability (imagine tooling for Support reps). I also didn't feel bad about exposing the `nlohmann::json` type, because it is declared in a short forwarding header (`json_fwd.hpp`) that mentions types from the standard library only: <https://github.com/nlohmann/json/blob/8c391e04fe4195d8be862c97f38cfe10e2a3472e/include/nlohmann/json_fwd.hpp#L37-L51>. Yes, that's a template template parameter.

PR #137, opened 2024-06-28
fix: trim inspected headers

dmehala:
## Description
Fixed a regression introduced in issue #42  that prevented HTTP headers with leading spaces from being parsed. This issue could result in the failure to extract trace context and the creation of a new trace, even though the trace could have been extracted and continued in versions prior to v0.1.11.


zacharycmontoya (approved):
LGTM

PR #142, opened 2024-07-19
feat: ensure tracecontext headers take precedence over datadog (AIT-10281)

zacharycmontoya:
# Description
Adds support for "W3C Phase 3" so that when multiple trace contexts are extracted (with the same trace-id) and the `tracecontext` format is used, the resulting trace context will have:
- `span.parent_id` set to the `tracecontext` parent-id value
- `span.tags["_dd.parent_id"]` set to either the value of `tracestate[dd][p]` or `hex(x-datadog-parent-id)`, in that order of precedence

### DIFF
--- a/src/datadog/extraction_util.cpp
+++ b/src/datadog/extraction_util.cpp
@@ -251,68 +251,44 @@ void AuditedReader::visit(
   });
 }
 
-ExtractedData merge(const std::vector<ExtractedData>& contexts) {
+ExtractedData merge(const PropagationStyle first_style,
+  const std::unordered_map<PropagationStyle, ExtractedData>& contexts) {
   ExtractedData result;
-
-  const auto found = std::find_if(
-      contexts.begin(), contexts.end(),
-      [](const ExtractedData& data) { return data.trace_id.has_value(); });
-
-  if (found == contexts.end()) {
-    // Nothing extracted a trace ID. Return the first context that includes a
-    // parent ID, if any, or otherwise just return an empty `ExtractedData`.
-    // The purpose of looking for a parent ID is to allow for the error
-    // "extracted a parent ID without a trace ID," if that's what happened.
-    const auto other = std::find_if(
-        contexts.begin(), contexts.end(),
-        [](const ExtractedData& data) { return data.parent_id.has_value(); });
-    if (other != contexts.end()) {
-      result = *other;
-    }
-    return result;
-  }
 
   // `found` refers to the first extracted context that yielded a trace ID.
   // This will be our main context.
   //
-  // If the style of `found` is not W3C, then examine the remaining contexts
-  // for W3C-style tracestate that we might want to include in `result`.
-  result = *found;
-  if (result.style == PropagationStyle::W3C) {
+  // If the W3C style is present and its trace-id matches, we'll update the main context
+  // with tracestate information that we want to include in `result`.
+  // We may also need to use Datadog header information (only when the trace-id matches).
+  const auto found = contexts.find(first_style);
+  if (found == contexts.end()) {
     return result;
   }
+  result = found->second;
 
-  const auto other =
-      std::find_if(found + 1, contexts.end(), [&](const ExtractedData& data) {
-        return data.style == PropagationStyle::W3C &&
-               data.trace_id == found->trace_id;
-      });
+  const auto w3c = contexts.find(PropagationStyle::W3C);
+  const auto dd = contexts.find(PropagationStyle::DATADOG);
 
-  const auto dd = std::find_if(
-      contexts.begin(), contexts.end(), [&](const ExtractedData& data) {
-        return data.style == PropagationStyle::DATADOG &&
-               data.trace_id == found->trace_id;
-      });
-
-  if (other != contexts.end()) {
-    result.additional_w3c_tracestate = other->additional_w3c_tracestate;
+  if (w3c != contexts.end() && w3c->second.trace_id == result.trace_id) {
+    result.additional_w3c_tracestate = w3c->second.additional_w3c_tracestate;
     result.additional_datadog_w3c_tracestate =
-        other->additional_datadog_w3c_tracestate;
+        w3c->second.additional_datadog_w3c_tracestate;
     result.headers_examined.insert(result.headers_examined.end(),
-                                   other->headers_examined.begin(),
-                                   other->headers_examined.end());
+                                   w3c->second.headers_examined.begin(),
+                                   w3c->second.headers_examined.end());
 
-    if (result.parent_id != other->parent_id) {
-      if (other->datadog_w3c_parent_id &&
-          other->datadog_w3c_parent_id != "0000000000000000") {
-        result.datadog_w3c_parent_id = other->datadog_w3c_parent_id;
-      } else if (dd != contexts.end() && dd->parent_id.has_value()) {
-        result.datadog_w3c_parent_id = hex_padded(dd->parent_id.value());
+    if (result.parent_id != w3c->second.parent_id) {
+      if (w3c->second.datadog_w3c_parent_id &&
+          w3c->second.datadog_w3c_parent_id != "0000000000000000") {
+        result.datadog_w3c_parent_id = w3c->second.datadog_w3c_parent_id;
+      } else if (dd != contexts.end() && dd->second.trace_id == result.trace_id && dd->second.parent_id.has_value()) {
+        result.datadog_w3c_parent_id = hex_padded(dd->second.parent_id.value());
         // TODO: Do we add the result.headers_examined for all of the dd ones?
         // Or just the x-datadog-parent-id?
       }
 
-      result.parent_id = other->parent_id;
+      result.parent_id = w3c->second.parent_id;
     }
   }
 
--- a/src/datadog/extraction_util.cpp
+++ b/src/datadog/extraction_util.cpp
@@ -288,8 +288,6 @@ ExtractedData merge(
                  dd->second.trace_id == result.trace_id &&
                  dd->second.parent_id.has_value()) {
         result.datadog_w3c_parent_id = hex_padded(dd->second.parent_id.value());
-        // TODO: Do we add the result.headers_examined for all of the dd ones?
-        // Or just the x-datadog-parent-id?
       }
 
       result.parent_id = w3c->second.parent_id;
--- a/src/datadog/extraction_util.cpp
+++ b/src/datadog/extraction_util.cpp
@@ -251,20 +251,22 @@ void AuditedReader::visit(
   });
 }
 
-ExtractedData merge(const PropagationStyle first_style,
-  const std::unordered_map<PropagationStyle, ExtractedData>& contexts) {
+ExtractedData merge(
+    const PropagationStyle first_style,
+    const std::unordered_map<PropagationStyle, ExtractedData>& contexts) {
   ExtractedData result;
-
-  // `found` refers to the first extracted context that yielded a trace ID.
-  // This will be our main context.
-  //
-  // If the W3C style is present and its trace-id matches, we'll update the main context
-  // with tracestate information that we want to include in `result`.
-  // We may also need to use Datadog header information (only when the trace-id matches).
   const auto found = contexts.find(first_style);
   if (found == contexts.end()) {
     return result;
   }
+
+  // `found` refers to the first extracted context that yielded a trace ID.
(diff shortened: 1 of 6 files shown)


dmehala (commented):
If my understanding of the RFC is correct, the sole addition of phase 3 is:
> MUST follow the rules below if `_dd.parent_id` tracestate if the `p` value is not set
> * Fallback 1: _dd.parent_id SHOULD be set using the parent_id extracted from datadog (x-datadog-parent-id). The value must be encoded according to W3C Phase 2 Plan
> * Fallback 2: Do not set _dd.parent_id

~~In that case, your change is unnecessary and only [this section of the code](https://github.com/DataDog/dd-trace-cpp/blob/f6d6836a765028fc4883676fae30456f78d9e354/src/datadog/tracer.cpp#L288-L291) needs to be updated~~.

EDIT: I forgot that the parent id must come from `x-datadog-parent-id`. Your change is still perfectible, when `tracecontext` is the first extraction style, and `p` is missing from the tracestate, then `_dd.parent_id` will not be set.

src/datadog/extraction_util.cpp
```diff
@@ -287,14 +288,32 @@ ExtractedData merge(const std::vector<ExtractedData>& contexts) {
...
     result.additional_w3c_tracestate = other->additional_w3c_tracestate;
     result.additional_datadog_w3c_tracestate =
         other->additional_datadog_w3c_tracestate;
     result.headers_examined.insert(result.headers_examined.end(),
                                    other->headers_examined.begin(),
                                    other->headers_examined.end());
+
+    if (result.parent_id != other->parent_id) {
+      if (other->datadog_w3c_parent_id &&
+          other->datadog_w3c_parent_id != "0000000000000000") {
```
dmehala, line 307: That condition was not needed before, why is it now?

zacharycmontoya: This condition that checks for value `"0000000000000000"` is needed now because if the `tracestate[dd][p]` value is not present or that default value, we must default to the Datadog `x-datadog-parent-id` value

zacharycmontoya:
> EDIT: I forgot that the parent id must come from `x-datadog-parent-id`. Your change is still perfectible, when `tracecontext` is the first extraction style, and `p` is missing from the tracestate, then `_dd.parent_id` will not be set.

In the case where `tracestate[dd][p]` is not present, the expected behavior is that `_dd.parent_id` will not be set, so my current implementation matches that. @mabdinur can you double-check that my interpretation of the requirement is correct? This implementation passes the system-tests so it seems like this is functioning as-expected.

> Searching for attribute X for propagation style Y seems inevitable, may I suggest a bigger change? Instead of pushing all extracted data to a `std::vector<ExtractedData>` [here](https://github.com/DataDog/dd-trace-cpp/blob/f6d6836a765028fc4883676fae30456f78d9e354/src/datadog/tracer.cpp#L153-L179), use an `std::unordered_map<PropagationStyle, ExtractedData>`.
> 
> Accessing an attributes for a given propagation style will be constant and improve the code readability. What do you think? :)

Yes this seems like a good change, I will follow up with a change to do that 👍🏼

src/datadog/extraction_util.cpp
zacharycmontoya: Resolving, but feel free to re-open if you have any further questions

src/datadog/extraction_util.cpp
```diff
@@ -250,51 +251,49 @@ void AuditedReader::visit(
...
+
+    if (result.parent_id != w3c->second.parent_id) {
+      if (w3c->second.datadog_w3c_parent_id &&
+          w3c->second.datadog_w3c_parent_id != "0000000000000000") {
+        result.datadog_w3c_parent_id = w3c->second.datadog_w3c_parent_id;
+      } else if (dd != contexts.end() &&
+                 dd->second.trace_id == result.trace_id &&
+                 dd->second.parent_id.has_value()) {
+        result.datadog_w3c_parent_id = hex_padded(dd->second.parent_id.value());
+        // TODO: Do we add the result.headers_examined for all of the dd ones?
```
zacharycmontoya, line 291: @dmehala what should I do here? I left this TODO to be finished and removed before PR merge but wanted your input here

dmehala:
> In the case where tracestate[dd][p] is not present, the expected behavior is that _dd.parent_id will not be set, so my current implementation matches that. @mabdinur can you double-check that my interpretation of the requirement is correct? This implementation passes the system-tests so it seems like this is functioning as-expected.

That's partially true. There's two _contradictory_ fallbacks. I'll let @mabdinur decide but my understanding is: set `_dd.parent_id` using the datadog parent header if available or do nothing. 

If we agree on the same understanding, `tracestate[dd][p]` will not be set while it should and that may be an indication of a missing sys-test.

zacharycmontoya:
Oh I think I understand where the confusion is, there's two different scenarios:
1. `tracecontext` is the first context
2. `tracecontext` is _not_ the first context

If `tracestate[dd][p]` is not set, do we use the `x-datadog-parent-id` as `_dd.parent_id` in both scenarios? Before my latest changes I only did this for Scenario 2. But now my changes should do this for Scenario 1 as well.

mabdinur:
> Oh I think I understand where the confusion is, there's two different scenarios:
> 
> 1. `tracecontext` is the first context
> 2. `tracecontext` is _not_ the first context
> 
> If `tracestate[dd][p]` is not set, do we use the `x-datadog-parent-id` as `_dd.parent_id` in both scenarios? Before my latest changes I only did this for Scenario 2. But now my changes should do this for Scenario 1 as well.

I updated the design over to state:

- MUST attempt to extract the last Datadog parent id from the tracestate headers (p value in dd=) regardless of the primary propagation style. The extracted value must be stored in a trace level tag with the key _dd.parent_id.
- If the tracestate header does not contain the last Datadog parent id AND the x-datadog-parent-id header is received, the value of the x-datadog-parent-id header must be converted to a 16 character hexidecimal and this value MUST be used to set the _dd.parent_id trace level tag.
- The span id from the transparent header MUST be used to continue the distributed trace. The span id extracted from traceparent headers should take precedence over ANY headers containing span ids, regardless of the primary propagation style.

zacharycmontoya:
> * If the tracestate header does not contain the last Datadog parent id AND the x-datadog-parent-id header is received, the value of the x-datadog-parent-id header must be converted to a 16 character hexidecimal and this value MUST be used to set the _dd.parent_id trace level tag.

I think what this is missing is defining the case where
1. Configured propagators are `tracecontext,Datadog`
2. The result is two context with the same trace-id but different span-ids
3. The context from `tracecontext` does not have the Last Datadog Parent Id in the tracestate

In this case, the RFC text says that we should use the x-datadog-parent-id value as the Last Datadog Parent Id (`meta[_dd.parent_id]'). Is this the desired outcome? The reason I ask is because other implementations and the pseudocode do not do this behavior

mabdinur:
> > * If the tracestate header does not contain the last Datadog parent id AND the x-datadog-parent-id header is received, the value of the x-datadog-parent-id header must be converted to a 16 character hexidecimal and this value MUST be used to set the _dd.parent_id trace level tag.
> 
> I am still confused on the following case:
> 
> 1. Configured propagators are `tracecontext,Datadog`
> 2. The result is two contexts with the same trace-id but different span-ids
> 3. The context from `tracecontext` does not have the Last Datadog Parent Id in the tracestate
> 
> In this case, the RFC text says that we should use the x-datadog-parent-id value as the Last Datadog Parent Id (`meta[_dd.parent_id]`). Is this the desired outcome? The reason I ask is because other implementations and the pseudocode do not do this behavior

That's correct. If `p` is not found in the tracestate then `x-datadog-parent-id` should be used to set `_dd.parent_id`. This is the sample implementation in the RFC:

```python
# when DD_TRACE_PROPAGATION_STYLE =datadog,tracecontext
# first_context is the SpanContext extracted from datadog headers
# dd_context is the SpanContext extracted from datadog headers
# w3c_context is the SpanContext extracted from w3c headers
if first_context.trace_id == w3c_context.trace_id and first_context.span_id != w3c_context.span_id:
   if "_dd.parent_id" in w3c_context._meta and w3c_context._meta["_dd.parent_id"] != "0000000000000000":
       # tracecontext headers contain a p value, ensure this value is sent to backend
       first_context._meta["_dd.parent_id"] = w3c_context._meta["_dd.parent_id"]
   elif dd_context:
       # if p value is not present in tracestate, use the parent id from the datadog headers
       first_context._meta["_dd.parent_id"] = "{:016x}".format(dd_context.span_id)
   # the span_id in tracecontext takes precedence over the first extracted propagation style
   first_context.span_id = w3c_context.span_id
```

We can ignore the case where `_dd.parent_id` is `0000000000000000` , this is deprecated logic and will be removed. 

- "_dd.parent_id" in w3c_context._meta:
    - In this scenario the `p` value was found and extracted from a traceparent header. The SpanContext with w3c trace information contains a valid `_dd.parent_id` tag. We want to copy this tag to the primary/first context.
 - dd_context
    - In this scenario trace information was successfully extracted from datadog headers and stored in a SpanContext. SpanContext.span_id stores the `x-datadog-parent-id` header. There is no need to accessed the `x-datadog-parent-id` header directly.

mabdinur:
> > * If the tracestate header does not contain the last Datadog parent id AND the x-datadog-parent-id header is received, the value of the x-datadog-parent-id header must be converted to a 16 character hexidecimal and this value MUST be used to set the _dd.parent_id trace level tag.
> 
> I am still confused on the following case:
> 
> 1. Configured propagators are `tracecontext,Datadog`
> 2. The result is two contexts with the same trace-id but different span-ids
> 3. The context from `tracecontext` does not have the Last Datadog Parent Id in the tracestate
> 
> In this case, the RFC text says that we should use the x-datadog-parent-id value as the Last Datadog Parent Id (`meta[_dd.parent_id]`). Is this the desired outcome? The reason I ask is because other implementations and the pseudocode do not do this behavior

We don't need to handle tracecontext,datadog. If the primary/first propagation style is tracecontext then we don't need to do any of the reparenting logic.

In that scenario the span_id and trace_id from the tracecontext headers are used and datadog headers are ignored. I can reaword the spec to make this clear

src/datadog/extraction_util.cpp
dmehala: `headers_examined` is mainly used for diagnostic purposes. Since this method is called `merge` I would argue the resulting `headers_examined` is the merge of all `headers_examined`.

The more I think about it, the more I think `headers_examined` should not be part `ExtractedData`. My advice would to not bother about it, a refactoring is needed.

src/datadog/tracer.cpp
```diff
@@ -175,11 +177,33 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
       return error->with_prefix(
           extraction_error_prefix(style, audited_reader.entries_found));
     }
-    extracted_contexts.push_back(std::move(*data));
-    extracted_contexts.back().headers_examined = audited_reader.entries_found;
+
+    if (!first_style_with_trace_id && (*data).trace_id.has_value()) {
```
dmehala, line 181:
```suggestion
    if (!first_style_with_trace_id && data->trace_id.has_value()) {
```

We usually use the `->`

src/datadog/tracer.cpp
```diff
@@ -175,11 +177,33 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
...
+      first_style_with_trace_id = style;
+    }
+
+    if (!first_style_with_parent_id && (*data).parent_id.has_value()) {
```
dmehala, line 185:
```suggestion
    if (!first_style_with_parent_id && data->parent_id.has_value()) {
```

src/datadog/tracer.cpp
```diff
@@ -175,11 +177,33 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
...
+      first_style_with_parent_id = style;
+    }
+
+    (*data).headers_examined = audited_reader.entries_found;
```
dmehala, line 189:
```suggestion
    data->headers_examined = audited_reader.entries_found;
```

dmehala (approved):
LGTM. Once you've replace `(*data).` for `data->`, it's good to merge. Thanks for contributing.

src/datadog/extraction_util.cpp
zacharycmontoya: Sounds good, thanks

khanayan123 (approved):

PR #143, opened 2024-08-06
Support changes for nginx-datadog/waf-rem-cfg

cataphract:
# Description

See individual commits

See https://github.com/DataDog/nginx-datadog/pull/71

# Motivation
The implemetation originally assumed that `service_target` was included in all configuration files. However,, this assumption is incorrect; for instance, Appsec configuration files does not include this field. This PR removes that unnecessary restriction.


dgoffredo:
Re: [this][1] commit...

It looks like Damien added `src/datadog` as an include directory to support his refactoring of remote config [here][2]. The structure of includes was no longer flat, so he chose to keep the `#include "..."` discipline flat by adding `src/datadog` to the compiler's list of include directories to search.

You then saw this causes problems with the POSIX header [glob.h][3] and so you removed `src/datadog` from the list of include directories (good idea) and started using relative includes instead, like `#include "../glob.h"`.

Here's an alternative workaround that I like better: Change the `#include` style throughout all of dd-trace-cpp to use angle-bracketed "qualified" includes. For example, in `src/datadog/remote_config/listener.h` you have `#include "../optional.h"`. Consider changing it instead to `#include <datadog/optional.h>`. In existing headers, everywhere there is `#include "foo.h"`, it would instead be `#include <datadog/foo.h>` or possibly `#include <datadog/remote_config/foo.h>`. The relative locations of the headers and sources that include each other would no longer be relevant -- it would always be relative to the exposed include directory `src/` and made unambiguous by the constant prefix `datadog/`.

This change would affect every file in the repository, but now that the structure of includes is not flat, I think it's The Right Way™.

The distinction really is just a matter of style. The advantage that comes to mind for what I propose is that someone interested in using the library programmatically can copy the include style without having to grok the examples. That and it's the style used at a previous job.

[1]: https://github.com/DataDog/dd-trace-cpp/pull/143/commits/c23c0ce849bbe01dbb4f48bf7f8735c66a8b1e7d
[2]: https://github.com/DataDog/dd-trace-cpp/commit/f3958f3a21f123e16ed042a00ea30e911514ca9f#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR208
[3]: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/glob.h.html

src/datadog/remote_config/remote_config.cpp
```diff
@@ -230,41 +240,47 @@ void Manager::process_response(const nlohmann::json& json) {
...
+
+        const auto& targeted_service = config_json.at("service_target");
+        if (targeted_service.at("service").get<StringView>() !=
+                tracer_signature_.default_service ||
+            targeted_service.at("env").get<StringView>() !=
+                tracer_signature_.default_environment) {
+          new_config.state = Configuration::State::error;
+          new_config.error_message = "Wrong service targeted";
+          goto skip_listeners;
+        }
```
dmehala, line 267: Does that mean outside of APM_TRACING product, other services do not target `service` and `env`?

cataphract: Remote config provides configuration only for the service/env that the remote config request specifies (client/client_tracer/{service,env})

cataphract:
> This change would affect every file in the repository, but now that the structure of includes is not flat, I think it's The Right Way™.
> 
> The distinction really is just a matter of style. The advantage that comes to mind for what I propose is that someone interested in using the library programmatically can copy the include style without having to grok the examples. That and it makes clearer that the file included is relative to some build-configured search directory, rather than being that _or_ relative to the current file (and not knowing which is the case).

@dgoffredo I'm not sold on this being a good idea, style aside because angled includes privilege include directories over local files (and hence you will have problems building another version of the library if you have another version installed globally), but anyway I've added a commit for it.

src/datadog/remote_config/remote_config.cpp
dmehala: The remote config request always include the service and env ([code for ref](https://github.com/DataDog/dd-trace-cpp/blob/aa037a9cdea70d50a193ff42173fa84038b1c24e/src/datadog/remote_config/remote_config.cpp#L107-L113). If understand correctly, then, this change is not required. Am I missing something?

dmehala:
@cataphract don't mind fixing the include issue. This issue is being addressed in https://github.com/DataDog/dd-trace-cpp/pull/144

src/datadog/remote_config/remote_config.cpp
cataphract: Yes, the **request** always includes the service and env, which is how the backend decides which configuration to send for asm. But this is about processing the **response**. In particular, this is assuming that the response is a json file and that it includes `service` and `env` keys. This is only true for APM_TRACING configuration.

dmehala: Ok. Pardon me, I want to make sure I understand. Does that mean for ASM, the response is not necessarily a json file?

cataphract: It is a json file, yes. But it doesn't have any `service_target` key. Plus, duplicativeat the RC protocol level, the configuration need not be json.

The best thing to do here would be to move the block to the APM_TRACING listener. Notethat parsing the content as json at this point is duplicative; the configuration passed to the listener is in the form of std::string. So the listener needs to parse the content as json again.

cataphract:
I've rebased this and updated nginx-datadog to match. I didn't use your PR in nginx-datadog, though; removing nlhomann::json is a lot more work than you have there.

src/datadog/remote_config/remote_config.cpp
```diff
@@ -30,24 +30,31 @@ constexpr std::array<uint8_t, sizeof(uint64_t)> capabilities_byte_array(
...
   static const std::regex path_reg(
-      "^(datadog/\\d+|employee)/([^/]+)/[^/]+/[^/]+$");
+      "^(datadog/\\d+|employee)/([^/]+)/([^/]+)/[^/]+$");
 
   std::cmatch match;
   if (!std::regex_match(config_path.data(),
                         config_path.data() + config_path.size(), match,
                         path_reg)) {
-    return nullopt;
+    return {};
```
dmehala, line 46: I believe this change might not be necessary. Would it be possible to revert it?

src/datadog/remote_config/remote_config.cpp
```diff
@@ -231,41 +240,48 @@ void Manager::process_response(const nlohmann::json& json) {
         return;
       }
 
-      auto decoded_config =
-          base64_decode(target_it.value().at("raw").get<StringView>());
-
-      const auto config_json = nlohmann::json::parse(decoded_config);
+      auto raw_data = target_it->at("raw").get<StringView>();
+      auto decoded_config = base64_decode(raw_data);
```
dmehala, line 244: ditto

src/datadog/remote_config/remote_config.cpp
```diff
@@ -231,41 +240,48 @@ void Manager::process_response(const nlohmann::json& json) {
...
+        const auto& targeted_service = config_json.at("service_target");
+        if (targeted_service.at("service").get<StringView>() !=
+                tracer_signature_.default_service ||
+            targeted_service.at("env").get<StringView>() !=
+                tracer_signature_.default_environment) {
+          new_config.state = Configuration::State::error;
+          new_config.error_message = "Wrong service targeted";
+          goto skip_listeners;
+        }
+      }
```
dmehala, line 269: TODO.
[genuine question] Are you planning to move the code to the listener?

src/datadog/tracer.cpp
```diff
@@ -83,15 +83,27 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
...
   // clang-format off
   auto config = nlohmann::json::object({
     {"version", tracer_version_string},
     {"runtime_id", runtime_id_.string()},
-    {"collector", collector_->config()},
+    {"collector", nlohmann::json::parse(collector_->config())},
     {"span_sampler", span_sampler_->config_json()},
-    {"injection_styles", join_propagation_styles(injection_styles_)},
-    {"extraction_styles", join_propagation_styles(extraction_styles_)},
+    {"injection_styles", to_string_vec(injection_styles_)},
```
dmehala, line 105: Can you confirm the change?
| Before | After |
| ------ | ----- |
|"tracecontext,datadog"|["tracecontext", "datadog"]| 

If so, thank you—this is indeed an improvement. However, the best approach would be to leverage the [JSON library](https://github.com/nlohmann/json?tab=readme-ov-file#basic-usage).

The code should look something like:
```cpp
void to_json(nlohmann::json& j, const PropagationStyle& style) {
   j = to_string_view(style);
}
```

And the usage:
```cpp
...
{"injection_styles", injection_styles_},
...
```

dmehala (commented):
Left a few comments. Thank you.

src/datadog/remote_config/remote_config.cpp
cataphract: This makes debugging easier.

include/datadog/logger.h
```diff
@@ -60,6 +60,9 @@ class Logger {
 
   virtual void log_error(const Error&);
   virtual void log_error(StringView);
+
+  virtual void log_debug(const LogFunc&);
+  virtual void log_debug(StringView);
```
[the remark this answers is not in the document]
cataphract: Done

src/datadog/remote_config/remote_config.cpp
dmehala: [question] Do you mind if I push a commit to remove that technical debt?

src/datadog/tracer.cpp
cataphract: Yes, that's the change. This also fixes the test `TestConfiguration.test_in_http` in nginx. With the changes made in this PR it now looks normal:

```json
{
  "collector": {
    "config": {
      "event_scheduler": {
        "type": "datadog::nginx::NgxEventScheduler"
      },
      "flush_interval_milliseconds": 2000,
      "http_client": {
        "type": "datadog::tracing::Curl"
      },
      "remote_configuration_url": "http://bogus:1234/v0.7/config",
      "request_timeout_milliseconds": 2000,
      "shutdown_timeout_milliseconds": 2000,
      "telemetry_url": "http://bogus:1234/telemetry/proxy/api/v2/apmtelemetry",
      "traces_url": "http://bogus:1234/v0.4/traces"
    },
    "type": "datadog::tracing::DatadogAgent"
  },
  "defaults": {
    "environment": "fooment",
    "service": "foosvc",
    "service_type": "web"
  },
  "environment_variables": {
    "DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS": "1"
  },
  "extraction_styles": [
    "B3",
    "Datadog"
  ],
  "injection_styles": [
    "B3",
    "Datadog"
  ],
  "report_traces": true,
  "runtime_id": "31998223-6dc3-4fd6-b4c2-cd1ded711143",
  "span_sampler": {
    "rules": []
  },
  "tags_header_size": 512,
  "trace_sampler": {
    "max_per_second": 200,
    "rules": []
  },
  "version": "[dd-trace-cpp version v0.2.2]"
}
```

dmehala: Indeed. Thanks, I appreciate.

src/datadog/remote_config/remote_config.cpp
cataphract: sure go ahead

dmehala (approved):
LGTM. Before merging, could you update the commit message to be more descriptive? Something like, "fix(remote_config): remove unnecessary restriction on configuration content" would be more appropriate.

src/datadog/remote_config/remote_config.cpp
dmehala: done in d11cd521edf61625bf49425de5ea0235a0b0b1d3

PR #149, opened 2024-08-26
patch some build/test scripts in `bin/`

dgoffredo:
I just set up VSCode's clangd extension for C++ source browsing, and was giving it a test drive in dd-trace-cpp.

`bin/test` was failing on my machine due to changes needed in the build configuration of `test/system-tests`, and `bin/check` was failing due to changes needed to `bin/bazel-build`. At least on my machine.

If some of the helper scripts in `bin/` are no longer in use, it might be best to delete them. For now, I propose these changes to get things working again locally on my machine.

### DIFF
--- a/bin/bazel-build
+++ b/bin/bazel-build
@@ -4,4 +4,13 @@ set -e
 
 cd "$(dirname "$0")"/..
 
-bazelisk build dd_trace_cpp
+if [ "$1" = '--absl' ]; then
+  rcfile=.bazelrc.absl
+elif [ "$1" = '--std' ]; then
+  rcfile=.bazelrc.std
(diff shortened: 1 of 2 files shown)


dmehala:
Wow it's already been 3 weeks. My apologies @dgoffredo. I recently updated CMakeLists.txt, can you try without the changes on `test/system-tests/CMakeLists.txt`?

dgoffredo:
I force pushed an update that removes the CMake changes and rebases onto `main`.

Keep in mind that the branch name "patch-system-tests-build" is now a misnomer, since this doesn't patch the system tests build anymore.

dmehala (approved):
LGTM! Thank you @dgoffredo

PR #151, opened 2024-08-29
fix: bugs in Remote Configuration payload construction

cataphract:
# Description

Fix bugs in remote config found during validation.

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/src/datadog/remote_config/remote_config.cpp
+++ b/src/datadog/remote_config/remote_config.cpp
@@ -154,12 +154,12 @@ nlohmann::json Manager::make_request_payload() {
       nlohmann::json cached_file = {
           {"path", config.path},
           {"length", config.content.size()},
-          {"hashes", {{"algorithm", "sha256"}, {"hash", config.hash}}}};
+          {"hashes", {{{"algorithm", "sha256"}, {"hash", config.hash}}}}};
 
       cached_target_files.emplace_back(std::move(cached_file));
     }
 
-    j["cached_target"] = cached_target_files;
+    j["cached_target_files"] = cached_target_files;
     j["client"]["state"]["config_states"] = config_states;
   }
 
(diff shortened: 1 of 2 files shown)


src/datadog/remote_config/remote_config.cpp
```diff
@@ -168,12 +168,26 @@ nlohmann::json Manager::make_request_payload() {
...
+
+    ~UpdateTargetsVersion() noexcept {
+      // if there was a global error, the old targets_version should be sent in
+      // the next request (at least this is what the system-tests expect)
+      if ((!state.error_message || state.error_message->empty()) &&
+          new_targets_version != std::uint64_t(-1)) {
+        state.targets_version = new_targets_version;
+      }
+    }
+  } update_targets_version{state_, std::uint64_t(-1)};
```
dmehala, line 183: This seems a bit complex, especially when it might be simpler to just update the field after the entire process is successful. Am I missing something here?

cataphract: that's basically it. It's just that the function has multiple returns. But it can be done other ways of course (goto, new wrapping function, more nested ifs, etc.)

dmehala: Hey @cataphract, your solution is clever, but the RC code is already quite complex (which is entirely my fault) and not easy to grasp. Adding this defer behavior would create unnecessary additional cognitive load. I took the liberty of pushing a commit that I believe is simpler and easier to understand.

test/remote_config/test_remote_config.cpp
```diff
@@ -268,6 +295,25 @@ REMOTE_CONFIG_TEST("response processing") {
...
       }
+
+      // Verify `cached_target_files` is reported
+      auto cached_target_files = payload.find("cached_target_files");
+      REQUIRE(cached_target_files != payload.end());
+      REQUIRE(cached_target_files->is_array());
+      REQUIRE(cached_target_files->size() == 3);
+
+      const auto ctf = cached_target_files->at(0);
+      CHECK(ctf.at("path").get<std::string_view>() ==
```
cataphract, line 306: tests are failing. You need a sort here

dmehala (approved):

test/remote_config/test_remote_config.cpp
dmehala: fixed in 4d15b7dd0a086b25c8bf6956edf42ee558895a3f

PR #157, opened 2024-10-03
fix: Apply "version" tag only when the span service name matches the default service name

zacharycmontoya:
# Description
Apply version tag to a span only if its service name matches the default service name. This may be viewed as a breaking change since now some spans that are started with a new service name will not have the "version" tag set.

# Motivation
This behavior conforms to our Config Consistency initiative

# Additional Notes
As this may be breaking, we could put this new behavior behind a feature flag, but I'm inclined to say that this is a FIX rather than a FEATURE.

Jira ticket: [APMAPI-517]

[APMAPI-517]: https://datadoghq.atlassian.net/browse/APMAPI-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/datadog/span_data.cpp
+++ b/src/datadog/span_data.cpp
@@ -37,7 +37,19 @@ Optional<StringView> SpanData::version() const {
 
 void SpanData::apply_config(const SpanDefaults& defaults,
                             const SpanConfig& config, const Clock& clock) {
-  service = config.service.value_or(defaults.service);
+  std::string version;
+  if (config.service) {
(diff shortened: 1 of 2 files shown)


dmehala (approved):
LGTM :shipit: I briefly updated the `span_data.cpp` to make the logic more explicit/easier to comprehend.

PR #158, opened 2024-10-03
feat!: Implement Config Consistency for DD_SERVICE when no service is configured

zacharycmontoya:
# Description
Adds a default service for the configuration if none was set in the `DD_SERVICE` environment variable or user configuration. This is a breaking change as previously a missing service would throw a configuration initialization error.

The default value will be set to the name of the process.

# Motivation
This aligns the cpp tracer behavior for an unconfigured `DD_SERVICE` with the expected behavior of all the tracers in our Config Consistency initiative.

# Additional Notes
Also updates the parametric test so that the DD_ENV behavior works as expected.

Jira ticket: [APMAPI-709] [APMAPI-708]


src/datadog/platform_util.cpp
```diff
@@ -189,6 +189,26 @@ int get_process_id() {
 #endif
 }
 
+std::string get_process_name() {
+#if defined(__APPLE__) || defined(__FreeBSD__)
+  return getprogname();
```
dmehala, line 194:
```suggestion
  char* process_name = getprogname();
  return (process_name != nullptr) ? process_name : "unknown-service";
```

`man getprogname`
> The getprogname() function returns the name of  the  program.   If  the
       name has	not been set yet, it will return NULL.

An `std::string` constructed with a NULL pointer will throw an exception.

src/datadog/platform_util.cpp
```diff
@@ -189,6 +189,26 @@ int get_process_id() {
...
+#elif defined(_MSC_VER)
+  TCHAR szExeFileName[MAX_PATH];
+  GetModuleFileName(NULL, szExeFileName, MAX_PATH);
+#ifdef UNICODE
+  std::wstring wStr(szExeFileName);
+  std::string path = std::string(wStr.begin(), wStr.end());
+#else
+  std::string path = std::string(szExeFileName);
+#endif
+  return path.substr(path.find_last_of("/\\") + 1);
```
dmehala, line 206:
```suggestion
  TCHAR exe_name[MAX_PATH];
  if (GetModuleFileName(NULL, exe_name, MAX_PATH) <= 0) {
    return "unknown-service";
  } 
#ifdef UNICODE
  std::wstring wStr(exe_name);
  std::string path = std::string(wStr.begin(), wStr.end());
#else
  std::string path = std::string(exe_name);
#endif
  return path.substr(path.find_last_of("/\\") + 1);
```

Changes:
- Renamed `szExeFileName` to `exe_name`.
- Check for `GetModuleFileName` returned value.

dmehala (approved):
Propose changes to consider cases where the underlying function can fail for Darwin and Windows NT.

PR #160, opened 2024-10-03
feat!: Change type of DD_TRACE_RATE_LIMIT to int

zacharycmontoya:
# Description
Changes the type of the `DD_TRACE_RATE_LIMIT` input to `int` (previously `double`)

# Motivation
This aligns the cpp tracer behavior for DD_TRACE_RATE_LIMIT with the expected behavior of all the tracers in our Config Consistency initiative, such that only integer values are used.

# Additional Notes
This may be unnecessary as we may be able to standardize our documentation that only integer values are accepted (what even is 0.1 of a trace?) and in doing so we are still compatible because integers on the command line can still be interpreted as a double. However, it could be good to do this in the type system just to be more specific. Thoughts?

Jira ticket: [APMAPI-511]

[APMAPI-511]: https://datadoghq.atlassian.net/browse/APMAPI-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/include/datadog/sampling_decision.h
+++ b/include/datadog/sampling_decision.h
@@ -38,7 +38,7 @@ struct SamplingDecision {
   Optional<Rate> limiter_effective_rate;
   // The per-second maximum allowed number of "keeps" configured for the limiter
   // consulted in this decision, if any.
-  Optional<double> limiter_max_per_second;
+  Optional<int> limiter_max_per_second;
   // The provenance of this decision.
   Origin origin;
 };
--- a/include/datadog/span_sampler_config.h
+++ b/include/datadog/span_sampler_config.h
@@ -23,7 +23,7 @@ namespace tracing {
 struct SpanSamplerConfig {
   struct Rule : public SpanMatcher {
     double sample_rate = 1.0;
-    Optional<double> max_per_second;
+    Optional<int> max_per_second;
 
     Rule(const SpanMatcher&);
     Rule() = default;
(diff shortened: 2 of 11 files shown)


dgoffredo:
> what even is 0.1 of a trace?

0.1 traces in a second is 1 trace in 10 seconds. `DD_TRACE_RATE_LIMIT` has units of "traces per second," not "traces."

Possibly best to make it an integer for consistency's sake, but I did [make sure][1] that the existing `Limiter` implementation could handle values smaller than 1.

[1]: https://github.com/DataDog/dd-trace-cpp/blame/85370e70703ac54990d2f903c128bbcb43d71f6f/test/test_limiter.cpp#L101-L124

zacharycmontoya:
> > what even is 0.1 of a trace?
> 
> 0.1 traces in a second is 1 trace in 10 seconds. `DD_TRACE_RATE_LIMIT` has units of "traces per second," not "traces."
> 
> Possibly best to make it an integer for consistency's sake, but I did [make sure](https://github.com/DataDog/dd-trace-cpp/blame/85370e70703ac54990d2f903c128bbcb43d71f6f/test/test_limiter.cpp#L101-L124) that the existing `Limiter` implementation could handle values smaller than 1.

Ah you're right since it's traces _per second_ it makes sense why this was implemented as a double earlier. It seems that the remaining languages only use integer values so there's a strong motivation to align. Thanks for clearing that up @dgoffredo 😄

dmehala:
Stale.

PR #161, opened 2024-10-08
Downgrade CMake version to 3.22.1

salmaanrizvi:
# Description
Downgrade CMake to 3.22.1 from 3.24.2

# Motivation
The latest version of CMake available for Ubuntu 22.04 via apt is [3.22.1](https://launchpad.net/ubuntu/jammy/+source/cmake). Use that version instead of requiring users to install a later version of CMake than may be available on their system.

# Additional Notes
I'm not sure if there are specific features from 3.23 or 3.24 that are required to support Windows or some other features, but I thought I would open a PR to at least start a discussion.

We had to fork the repository internally to downgrade so I thought it was worth committing back upstream to see if this was a change that would be accepted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,10 +82,7 @@ file(GLOB_RECURSE public_header_files ${CMAKE_SOURCE_DIR}/include/*.h)
 
 target_sources(dd_trace_cpp-objects
   PUBLIC
-    FILE_SET public_headers
-    TYPE HEADERS
-    BASE_DIRS include
-    FILES ${public_header_files}
+    ${public_header_files}
   PRIVATE
     src/datadog/base64.cpp
     src/datadog/cerr_logger.cpp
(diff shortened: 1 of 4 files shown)


dmehala:
Hey @salmaanrizvi 

Thank you so much for taking the time to open this PR and contribute to the project. After reviewing the proposed changes, we won’t be able to proceed with downgrading CMake. While I understand the motivation to align with the default version provided by Ubuntu 22.04, it is not a valid reason to downgrade.

We’ve provided [a script](https://github.com/DataDog/dd-trace-cpp/blob/36d1b602f996c83e3a056bdfa9c7067b5f3f2a38/bin/install-cmake) to handle this exact situation — it ensures that anyone working on the project can easily install the required version of CMake, even if their system’s default version is lower. I would recommend running the script to ensure compatibility on your end.

Again, I really appreciate your contribution. Feel free to reach out if you have any other questions or ideas.

PR #169, opened 2024-11-15
build: support `clang-cl` compiler

dmehala:
# Description
In another internal project that relies on `dd-trace-cpp`, I needed to enable cross-compilation for Windows from a Linux host. During this process, I used `clang-cl` as the compiler - MSVC frontend for `clang`. However, I encountered an issue where `CMAKE_CXX_COMPILER_ID` was detected as `Clang`, which caused the incorrect file to be included, leading to a compilation failure.

To address this, I utilized the `CMAKE_CXX_COMPILER_FRONTEND_VARIANT` variable. This change ensures that the correct compiler frontend (`clang-cl`) is identified, allowing for successful cross-compilation in this scenario.


pablomartinezbernardo (approved):
Thanks for the explanation!

PR #170, opened 2024-11-15
testing(parametric): update start span parametric endpoint

mabdinur:
# Description

- Moves http_header extraction from `/trace/span/start/` to a new `trace/span/extract_headers` endpoint. This ensures the `/trace/span/start` parametric endpoint aligns with the following [specification](https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_parametric_endpoints.py#L25).
- Ensures `/trace/span/finish` does NOT clear the span list. Spans should only be removed from `spans_` (previously `active_spans_`) if the `trace/spans/flush` is called.
-  R̶e̶n̶a̶m̶e̶s̶ ̶t̶h̶e̶ ̶`̶a̶c̶t̶i̶v̶e̶_̶s̶p̶a̶n̶s̶_̶`̶ ̶v̶a̶r̶i̶a̶b̶l̶e̶ ̶t̶o̶ ̶`̶s̶p̶a̶n̶s̶_̶`̶.̶ ̶T̶h̶i̶s̶ ̶m̶a̶p̶ ̶w̶i̶l̶l̶ ̶n̶o̶ ̶l̶o̶n̶g̶e̶r̶ ̶j̶u̶s̶t̶ ̶c̶o̶n̶t̶a̶i̶n̶ ̶a̶c̶t̶i̶v̶e̶ ̶s̶p̶a̶n̶s̶,̶ ̶i̶t̶ ̶w̶i̶l̶l̶ ̶a̶l̶s̶o̶ ̶s̶t̶o̶r̶e̶ ̶s̶p̶a̶n̶s̶ ̶e̶x̶t̶r̶a̶c̶t̶e̶d̶ ̶f̶r̶o̶m̶ ̶d̶i̶s̶t̶r̶i̶b̶u̶t̶e̶d̶ ̶t̶r̶a̶c̶i̶n̶g̶ ̶h̶e̶a̶d̶e̶r̶s̶.̶

# Motivation

- Follow up to: https://github.com/DataDog/system-tests/pull/3300/commits/890bccdc369a20b6057546d2fff6ae6d6fd64652
- After this PR is merged we can revert the changes in `manifest/cpp.yaml` that were introduced in https://github.com/DataDog/system-tests/pull/3300

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/test/system-tests/request_handler.cpp
+++ b/test/system-tests/request_handler.cpp
@@ -230,6 +230,7 @@ void RequestHandler::on_extract_headers(const httplib::Request& req,
     VALIDATION_ERROR(res, "on_extract_headers: missing `http_headers` field.");
   }
 
+  datadog::tracing::SpanConfig span_cfg;
   auto span = tracer_.extract_span(
       utils::HeaderReader(*http_headers), span_cfg);
 
--- a/test/system-tests/request_handler.cpp
+++ b/test/system-tests/request_handler.cpp
@@ -257,6 +257,8 @@ void RequestHandler::on_extract_headers(const httplib::Request& req,
 
   res.set_content(response_body.dump(), "application/json");
   http_headers_.emplace(span->parent_id().value(), std::move(*http_headers));
+  // The span below will not be finished and flushed.
+  extract_headers_spans_.emplace(span->parent_id().value(), std::move(*span));
 }
 
 void RequestHandler::on_span_flush(const httplib::Request& /* req */,
--- a/test/system-tests/request_handler.cpp
+++ b/test/system-tests/request_handler.cpp
@@ -256,9 +256,9 @@ void RequestHandler::on_extract_headers(const httplib::Request& req,
   // clang-format on
 
   res.set_content(response_body.dump(), "application/json");
-  http_headers_.emplace(span->parent_id().value(), std::move(*http_headers));
+  http_headers_[span->parent_id().value()] = std::move(*http_headers);
   // The span below will not be finished and flushed.
-  extract_headers_spans_.emplace(span->parent_id().value(), std::move(*span));
+  extract_headers_spans_.push_back(std::move(*span));
 }
 
 void RequestHandler::on_span_flush(const httplib::Request& /* req */,
--- a/test/system-tests/request_handler.cpp
+++ b/test/system-tests/request_handler.cpp
@@ -221,34 +221,29 @@ void RequestHandler::on_inject_headers(const httplib::Request& req,
   res.set_content(response_json.dump(), "application/json");
 }
 
-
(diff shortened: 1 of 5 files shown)


test/system-tests/request_handler.cpp
```diff
@@ -240,9 +221,34 @@ void RequestHandler::on_inject_headers(const httplib::Request& req,
...
+
+  datadog::tracing::SpanConfig span_cfg;
+  auto span =
+      tracer_.extract_span(utils::HeaderReader(*http_headers), span_cfg);
+  if (auto error = span.if_error()) {
+    VALIDATION_ERROR(res, error->with_prefix("on_extract_headers: ").message);
+  }
+
+  const auto response_body = nlohmann::json{
+      {"span_id", span->id()},
```
mabdinur, line 241:
```suggestion
      {"span_id", span->parent_id()},
```

test/system-tests/request_handler.cpp
```diff
@@ -240,9 +221,34 @@ void RequestHandler::on_inject_headers(const httplib::Request& req,
...
+  const auto request_json = nlohmann::json::parse(req.body);
+  auto http_headers = utils::get_if_exists<nlohmann::json::array_t>(
+      request_json, "http_headers");
+  if (!http_headers) {
+    VALIDATION_ERROR(res, "on_extract_headers: missing `http_headers` field.");
+  }
+
+  datadog::tracing::SpanConfig span_cfg;
+  auto span =
+      tracer_.extract_span(utils::HeaderReader(*http_headers), span_cfg);
```
mabdinur, line 235: This operation creates a child span from distributed tracing headers. This is not what we need. 

We need a way to extract distributed tracing headers and store it in a "SpanContext" or a local representation of the remote span (ex: MergedContext or a "RemoteSpan"). 

To achieve this functionality we need to modify the cpp tracer (ie add an extract_headers(...) function to the public API). The cpp tracer is the only tracer that does not provide this API.

test/system-tests/request_handler.cpp
mabdinur, line 241: Here we want to return the span_id extracted from the distributed tracing headers not the span id of the newly created span.

```suggestion
      {"span_id", span->parent_id()},
```

test/system-tests/request_handler.cpp
```diff
@@ -6,6 +6,7 @@
 #include <datadog/tracer_config.h>
 
 #include <datadog/json.hpp>
+#include <type_traits>
```
dmehala, line 9:
```suggestion
```

dmehala (approved):
LGTM. I changed a bit the logic to avoid the endless if/else cascade and documented a bit the change.

PR #172, opened 2024-12-01
Properly support setting DD_TRACE_TRANSPORT to "none"

kristjanvalur:
# Description

Omit the 'curl' library if DD_TRACE_TRANSPORT is set to "none"

# Motivation

There is clearly the intent to be able to not include a default transport, by setting `DD_TRACE_TRANSPORT=none`.
However, two things were missing:

1. To ignore the bundled curl library from the build
2. to replace the `default_http_client_curl.cpp` with `default_http_client_null.cpp`

Integrations that provide their own transport may not want to build and link a whole unused library, along with any runtime library dependencies it makes.

# Additional Notes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,15 @@ endif ()
 
 set(DD_TRACE_TRANSPORT "curl" CACHE STRING "HTTP transport that dd-trace-cpp uses to communicate with the Datadog Agent, can be either 'none' or 'curl'")
 
+if(DD_TRACE_TRANSPORT STREQUAL "curl")
+  include(cmake/deps/curl.cmake)
+  message(STATUS "DD_TRACE_TRANSPORT is set to 'curl', including curl")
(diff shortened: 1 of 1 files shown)


dmehala:
Fantastic @kristjanvalur.
This is indeed the intended behaviour. I appreciate your contributions.

dmehala (approved):
LGTM!
If I recall correctly, when no HTTP client is provided, the library defaults to `default_http_client`. This could _potentially_ lead to a crash since the library assumes `default_http_client` is always a valid instance. While it functions, some refactoring would be beneficial to help users avoid unintended issues.

I think the tradeoff is ok, as long as it help you progress on your plugin :)

Thank you!

kristjanvalur:
I could add a check in the config builder for a nullptr, but since this is an integration issue and not a configuration issue, in my view assuming a non-null ptr (and thus a correctly written integration) is ok.

PR #177, opened 2025-01-21
fix: Remote config service/env validation

zacharycmontoya:
# Description
Removes the remote config check that asserts a matching env / service, as this is incorrectly causing remote config payloads to be ignored

# Motivation
This was motivated by a customer issue with adaptive sampling not working. When the tracer doesn't report an environment version but the Datadog agent does, incoming remote config payloads (which have already been verified by the Datadog agent) get rejected even though they are correct. This removes the validation and relies upon the Datadog agent verification.

# Testing
The previous unit test has been removed as it's invalid, and the associated system-tests authored here have been tested locally with success: https://github.com/DataDog/system-tests/pull/3778

# Additional Notes

Jira ticket: [APMAPI-1006]

[APMAPI-1006]: https://datadoghq.atlassian.net/browse/APMAPI-1006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/datadog/config_manager.cpp
+++ b/src/datadog/config_manager.cpp
@@ -151,14 +151,6 @@ Optional<std::string> ConfigManager::on_update(const Configuration& config) {
 
   const auto config_json = nlohmann::json::parse(config.content);
 
-  const auto& targeted_service = config_json.at("service_target");
-  if (targeted_service.at("service").get<StringView>() !=
-          tracer_signature_.default_service ||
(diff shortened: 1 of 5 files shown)


dmehala (approved):
LGTM! :shipit:

PR #179, opened 2025-02-03
feat: add baggage API

dmehala:
# Description
Add a baggage API.

# Todo:
- [ ] Memory allocation optimization.
- [x] Strict parsing (restrict characters allowed).

# Additional Notes
Jira ticket: [APMAPI-196]

[APMAPI-196]: https://datadoghq.atlassian.net/browse/APMAPI-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


examples/baggage/main.cpp
dmehala, line 1: add `get_all` method

include/datadog/baggage.h
```diff
@@ -0,0 +1,143 @@
...
+
+  /// Injects the Baggage data into a `DictWriter` with the constraint that
+  /// the amount of bytes written does not exceed the specified maximum byte
+  /// limit.
+  ///
+  /// @param `writer` The DictWriter to inject the data into.
+  /// @param `max_bytes` The maximum number of bytes to write.
+  /// @return An `Expected<void>`, which may either succeed or contain an
+  /// error.
+  Expected<void> inject(DictWriter& writer, size_t max_bytes) const;
```
dmehala, line 130: - [ ] Use `Expected<void, Baggage::Error>`

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+#include <cctype>
+
+namespace datadog {
+namespace tracing {
+
+namespace {
+
+Expected<std::unordered_map<std::string, std::string>, Baggage::Error>
+parse_baggage(StringView input, size_t max_capacity) {
```
dmehala, line 11: - [ ] Handle correctly `key=`. Return malformed baggage.

src/datadog/trace_segment.cpp
```diff
@@ -435,7 +435,6 @@ bool TraceSegment::inject(DictWriter& writer, const SpanData& span,
                               additional_w3c_tracestate_));
         break;
       default:
```
dmehala, line 437: - [ ] reset and continue on `PropagationStyle::BAGGAGE`.

src/datadog/tracer.cpp
```diff
@@ -80,6 +82,20 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
...
+
+  for (const auto style : injection_styles_) {
+    if (style == PropagationStyle::BAGGAGE) {
+      baggage_injection_enabled_ = true;
+      break;
+    }
+  }
+
   if (config.log_on_startup) {
     logger_->log_startup([configuration = this->config()](std::ostream& log) {
```
dmehala, line 100: - [ ] Log baggage config.

src/datadog/tracer.cpp
```diff
@@ -363,5 +378,34 @@ Span Tracer::extract_or_create_span(const DictReader& reader,
...
+  if (maybe_baggage) {
+    return std::move(*maybe_baggage);
+  }
+
+  return create_baggage();
+}
+
+Expected<void> Tracer::inject(const Baggage& baggage, DictWriter& writer) {
+  if (!baggage_injection_enabled_) {
+    // TODO(@dmehala): update `Expected` to support `<void, Error>`
```
dmehala, line 403: - [ ] TODO

test/test_baggage.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  CHECK(maybe_baggage->size() == 3);
+
+  CHECK(maybe_baggage->get("company") == "datadog");
+  CHECK(!maybe_baggage->get("boogaloo"));
+  CHECK(maybe_baggage->contains("boogaloo") == false);
+  CHECK(maybe_baggage->contains("team") == true);
+
+  maybe_baggage->set("color", "red");
+  /// NOTE: assure `set` overwrite
```
dmehala, line 194:
```suggestion
  /// NOTE: ensure `set` overwrite
```

fuzz/tracing/baggage.cpp
zacharycmontoya, line 1: nit: I think it would make more sense to name the folder `baggage` instead of `tracing` since in theory it's independent from tracing, it falls more under the umbrella of context propagation

include/datadog/baggage.h
```diff
@@ -0,0 +1,143 @@
...
+/// Baggage is a key-value store meant to propagate data across services and
+/// processes boundaries.
+///
+/// Baggage are extracted from any tracing context implementing the `DictReader`
+/// interface using `Baggage::extract`.
+///
+/// Baggages are injected to any tracing context implementing the `DictWriter`
+/// interface using the `inject` method.
+class Baggage {
+ public:
```
zacharycmontoya, line 24: The public API seems a little too large to me. Comparing this to the https://github.com/open-telemetry/opentelemetry-cpp/blob/5bc053e52280281736a4a07aec73d4f5f3500425/api/include/opentelemetry/baggage/baggage.h, can we remove the following API's from public accessibility:
- `contains`
- `size`
- `empty`
- `visit`

include/datadog/baggage.h
zacharycmontoya, line 130: What do you think of changing the `size_t max_bytes` argument into a `BaggageOptions` (you can choose the name) struct similar to `datadog::tracing::InjectionOptions`?

test/test_baggage.cpp
```diff
@@ -0,0 +1,214 @@
...
+        {
+            "spaces in key is allowed",
+            "key1 foo=value1",
+            Baggage({{"key1 foo", "value1"}}),
+        },
+        {
+            "verify separator",
+            "key1=value1;a=b,key2=value2",
+            Baggage({{"key1", "value1;a=b"}, {"key2", "value2"}}),
+        },
```
zacharycmontoya, line 105: These are both great edge cases, can you make sure these get included in our baggage system-tests? The current update is being made here: https://github.com/DataDog/system-tests/pull/3929

test/test_baggage.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+BAGGAGE_TEST("extract") {
+  SECTION("limit is respected") {
+    const std::unordered_map<std::string, std::string> headers{
+        {"baggage", "team=proxy,company=datadog,user=dmehala"}};
+    MockDictReader reader(headers);
+
+    auto extraction_fails = Baggage::extract(reader, 1);
+    REQUIRE(!extraction_fails);
+    CHECK(extraction_fails.error() == Baggage::Error::MAXIMUM_CAPACITY_REACHED);
```
zacharycmontoya, line 25: According to our Baggage spec, we seem to only enforce baggage limits on inject and not at the time of extract so I think this should move this error to the injection logic

test/test_baggage.cpp
```diff
@@ -0,0 +1,214 @@
...
+               test_case.expected_baggage.if_error()) {
+      CHECK(maybe_baggage.error() == test_case.expected_baggage.error());
+    } else {
+      FAIL("mistmatch between what is expected and the result");
+    }
+  }
+}
+
+BAGGAGE_TEST("inject") {
+  SECTION("limit is respected") {
```
zacharycmontoya, line 148: Can you also unit-test that the default max_bytes (8192) is respected?

test/test_baggage.cpp
```diff
@@ -0,0 +1,214 @@
...
+    Baggage bag({{"violets", "blue"}, {"roses", "red"}});
```
zacharycmontoya, line 149: Can you also test cases like this where injection will properly URL encode characters? I suppose this also applies to parsing as well.

Here's a round-trip example: https://github.com/DataDog/system-tests/blob/68fb7b7f5adfd47e7c02eb3b34a86c6fe671f8e0/tests/parametric/test_headers_baggage.py#L104

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,177 @@
...
+    res += "=";
+    res += it->second;
+  }
+
+  if (res.size() >= max_bytes)
+    return datadog::tracing::Error{
+        datadog::tracing::Error::Code::BAGGAGE_MAXIMUM_BYTES_REACHED, ""};
+
+  /// NOTE(@dmehala): It is the writer's responsibility to write the header,
+  /// including percent-encoding.
```
zacharycmontoya, line 154: I disagree with this conclusion and think it should be the responsibility of the injection operation to do percent-encoding, since baggage values can be any vali UTF-8 string

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  StringView key;
+  StringView value;
+
+  const size_t end = input.size();
+
+  for (size_t i = 0; i < end; ++i) {
+    switch (internal_state) {
+      case state::leading_spaces_key: {
+        if (input[i] != ' ') {
```
zacharycmontoya, line 31: I think we also need to check for `\t`

```suggestion
        if (input[i] != ' ' && input[i] != '\t') {
```

zacharycmontoya: And I think this can also be applied elsewhere. I'm not sure how efficient it is, but maybe you could define a lambda expression in this function that evaluates whether a char is optional whitespace

test/test_baggage.cpp
zacharycmontoya: Actually, based on further research, it does _not_ look like spaces are allowed in keys. Here's how I arrived this conclusion:
- The W3C spec says the [key](https://www.w3.org/TR/baggage/#key) must be a token defined in RFC7230, Section 3.2.6.
- And [RFC7230, Section 3.2.6](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) defines a token in the following way, where `VCHAR` is the only category of characters that a space may be:

```
     token          = 1*tchar

     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```

- In [RFC7230, Section 1.2](https://datatracker.ietf.org/doc/html/rfc7230#section-1.2) (same RFC) it states that the definition of `VCHAR` is referenced from RFC 5234, Appendix B.1
- In [RFC 5234, Appendix B.1](https://datatracker.ietf.org/doc/html/rfc5234#appendix-B.1), it defines `VCHAR` as the range `%x21-7E`, which does not include space `%x20`

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,177 @@
...
+  result.emplace(std::string(key), std::string(value));
+
+  return result;
+}
+
+}  // namespace
+
+Baggage::Baggage(size_t max_capacity) : max_capacity_(max_capacity) {}
+
+Baggage::Baggage(
```
zacharycmontoya, line 88: For the constructors, what do you think is the minimum set of constructors that could address the vast majority of use-cases? Could we get by with only the `std::unordered_map<std::string, std::string>` constructor? Or would it be possible to use a function template to make this generic, similar to what the OpenTelemetry API for C++ has done for [one](https://github.com/open-telemetry/opentelemetry-cpp/blob/5bc053e52280281736a4a07aec73d4f5f3500425/api/include/opentelemetry/baggage/baggage.h#L32-L33) of their constructors? I think it would be great to do something like that, but then we will likely need to allocate a new data structure rather than re-use the one provided by the user.

The reason I'm asking is because I want to limit the public API as much as reasonably possible, so I was thinking we could either have one constructor with a concrete type or we could

zacharycmontoya (commented):
I took a first look and the design looks good to me, but I have some open questions on some implementation details.

test/test_baggage.cpp
dmehala: Amazing! Akshually that makes everything simpler. Thank you @zacharycmontoya!

src/datadog/baggage.cpp
dmehala: I have to disagree. URL encoding is primarily relevant for HTTP propagation. Since this library isn't limited to HTTP, handling encoding should fall under the responsibility of the `DictWriter`/`DictReader`.  

For instance, if two processes communicate via IPC, they can define a serialization format that supports UTF-8 strings directly—eliminating the need for percent encoding altogether.

test/test_baggage.cpp
dmehala: Replied in https://github.com/DataDog/dd-trace-cpp/pull/179/files#r1947301140

src/datadog/baggage.cpp
dmehala: `std::unordered_map<std::string, std::string>` is the absolute minimum requirement. I reject any function template suggestions unless there's a concrete use case—"it could be useful" isn't a good enough reason.

include/datadog/baggage.h
dmehala: Unless there is a need, I don't think so.

include/datadog/expected.h
```diff
@@ -45,7 +45,7 @@
 namespace datadog {
 namespace tracing {
 
-template <typename Value>
+template <typename Value, typename Error = Error>
```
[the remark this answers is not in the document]
dmehala: > by default if only Expected<T> is used then we default to the datadog::tracing::Error type? 

Yes.

> And I'm guessing this is done so we can easily use the new datadog::tracing::Baggage::Error type?

Not really, it's mostly to keep the current behaviour (and avoiding to change the whole codebase).

include/datadog/baggage.h
dmehala: Is there a specific reason we should limit ourselves to OpenTelemetry's API? Just for context, `dd-trace-cpp` isn’t designed to be OpenTelemetry-compatible and doesn’t aim to be. Ultimately, if a user wants to stay vendor-neutral, they’ll likely choose 🚧 `dd-opentelemetry-cpp` 🚧.

zacharycmontoya: There's two reasons why I think we should generally strive to match the OpenTelemetry API:

1. If our users may switch from using the OpenTelemetry API to the API in our library (I guess maybe not the case for this repo), then closely matching the structure can help make it easier for users to adopt our library.
2. In general I think we should try to provide the most minimal API we can (with the ability to add to it later as needed), and the OpenTelemetry API's have been discussed enough that they should be a good starting point for us to reference as well.

So if we can get by without the additional class members public, I would greatly prefer that. But if you think they are necessary then we can keep them, it's not a blocker.

include/datadog/baggage.h
zacharycmontoya: I also don't see a direct need today, but I was thinking that it would be a better user that if we needed to add options then we could keep the function signature the same and just add to an options class. Anyways, this concern isn't a blocker

src/datadog/baggage.cpp
zacharycmontoya: Understood, I have no concrete use cases so we won't do a function template 👍🏼

And I'm just now learning about `std::initializer_list`. While it seems handy, I would prefer leaving it out so the only constructor that has existing baggage items is the `std::unordered_map<std::string, std::string>` one

src/datadog/baggage.cpp
zacharycmontoya: Ah ok, I see your point, so that responsibility should not be done here.

One concern I still have, though, is thinking about the experience this as a library developer from another programming language like C#: If I'm a user and I am adding baggage using the Baggage API, I expect that I can use any characters that are valid in UTF-8 strings and then if I make an HTTP request, the tracing component (configured to propagate baggage) will automatically percent-encode the baggage and then add the baggage as an HTTP header.

I guess my question is, should end-users of this library expect the same experience? Or are they responsible for both performing the encoding step and storing baggage on downstream HTTP requests?

dmehala: Library users are responsible for setting HTTP headers, whether for tracing context or baggage. This responsibility includes encoding and decoding headers, although this task is typically managed by the underlying web framework. For instance, you can see how this is handled in the `nginx-datadog` project [here](https://github.com/DataDog/nginx-datadog/blob/f1e5afffa7338767750d51ee879fe4a6ce4e28e8/src/ngx_header_writer.h#L17).

Unlink some other languages, there is no standard interface or de factor HTTP library for cpp. Different libraries and frameworks manage HTTP headers in various ways. For example, some store headers as `std::unordered_map<std::string, std::string>`, while others use `std::vector<Headers>`. Certain frameworks/intergrations, such as NGINX, employ their own data models optimized for their specific needs.
Rather than imposing our own interpretation of how HTTP headers should be stored, which could confuse users, we provide the flexibility for users to manage headers in the way that best suits their requirements. We simply reauire the implementation of a `DictReader`/`DictWriter` interface that will retrieve the necessary headers in their datamodel.

TL;DR: propagation over HTTP is not automatic and library user certainly don't have the same experience as C# library users. They are responsible for performing encoding/decoding and storage of HTTP headers.

zacharycmontoya: Got it, thank you for explaining. I'm on the same page as you now, so we can go ahead and close this 👍🏼

src/datadog/baggage.cpp
dmehala: Addressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

src/datadog/tracer.cpp
dmehala: Will be addressed in another PR.

include/datadog/baggage.h
dmehala: Adressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

test/test_baggage.cpp
dmehala: Adressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

test/test_baggage.cpp
dmehala: Adressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

test/test_baggage.cpp
dmehala: Adressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

src/datadog/baggage.cpp
dmehala: Adressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

src/datadog/baggage.cpp
dmehala: Adressed in e3394c9f9e776b0f769c160914b30aeda6dea13a

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,250 @@
...
+  for (const auto& [key, value] : baggage_) {
+    visitor(key, value);
+  }
+}
+
+Expected<void> Baggage::inject(DictWriter& writer, const Options& opts) const {
+  if (baggage_.empty()) return {};
+  if (baggage_.size() > opts.max_items)
+    return datadog::tracing::Error{
+        datadog::tracing::Error::Code::BAGGAGE_MAXIMUM_BYTES_REACHED, ""};
```
zacharycmontoya, line 205: Our RFC states that when we exceed the configured item/bytes limits upon injection, the the APM SDKs should drop baggage name/value pairs until under the limit (and then continue to send the new value). Can you double-check with @lucaspimentel or @rachelyangdog to confirm whether this really is optional (SHOULD) or whether this behavior is required (MUST)?

zacharycmontoya, line 205: If we keep this behavior after the discussion posted above, then:

```suggestion
        datadog::tracing::Error::Code::MAXIMUM_CAPACITY_REACHED, ""};
```

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,250 @@
...
+  for (it++; it != baggage_.cend(); ++it) {
+    res += ",";
+    res += it->first;
+    res += "=";
+    res += it->second;
+  }
+
+  if (res.size() >= opts.max_bytes)
+    return datadog::tracing::Error{
+        datadog::tracing::Error::Code::BAGGAGE_MAXIMUM_BYTES_REACHED, ""};
```
zacharycmontoya, line 225: If we keep this behavior after the discussion posted above, then:

```suggestion
        datadog::tracing::Error::Code::MAXIMUM_BYTES_REACHED, ""};
```

src/datadog/baggage.cpp
```diff
@@ -0,0 +1,250 @@
...
+        } else if (is_whitespace(c)) {
+          tmp_end = i;
+          internal_state = state::trailing_spaces_value;
+        } else if (!is_allowed_value_char(c)) {
+          return Baggage::Error{Baggage::Error::MALFORMED_BAGGAGE_HEADER, i};
+        }
+      } break;
+
+      case state::trailing_spaces_value: {
+        if (c == ',') {
```
zacharycmontoya, line 128: I'm noticing that this algorithm doesn't account for additional `property` items in baggage, as described in the format [here](https://www.w3.org/TR/baggage/#definition). That is, if there is a baggage key-value pair, _then_ OWS, _then_ a `;` to begin a list of properties, it will be considered malformed. While I do not advocate that we must support this, we should make sure that we are handling this consistently across tracers and author a corresponding system-tests case.

@lucaspimentel @rachelyangdog has this been clearly documented in our system-tests?

zacharycmontoya (commented):
Overall this looks good but I would like some confirmation on some edge-cases I referenced in new comment threads.

Once you confirm that the implementation is in line with the expected cross-tracer expectations on these edge-cases, I'll be happy to approve. Thanks!

src/datadog/baggage.cpp
lucaspimentel: [The W3C spec](https://www.w3.org/TR/baggage/#limits) says:
- tracer MUST send all items if both conditions are met (max number of items, and max size in bytes)
- if either condition it not met, tracer MAY drop items until conditions are met (which items are dropped or in which order is undefined)
- only propagate complete items, never propagate partial items (e.g. key without value)

There is no mention of dropping the entire header, and I don't think we should ever to that.

src/datadog/baggage.cpp
lucaspimentel: We decided not to support properties (aka metadata) yet since OpenTelemetry themeselves have not fully adopted it. For example, the metadata parameter in the `Set` API is optional in the spec. Some implementations of the OpenTelemetry API support properties, but others don't. See https://github.com/open-telemetry/opentelemetry-specification/issues/3972#issuecomment-2031325588.

We should've been more clear about this in our RFC. If we won't support properties yet, we should probably ignoring them (not extract them). If we ignore them, for a baggage header like this:
```
baggage: key1=value1;property1;property2, key2 = value2, key3=value3; propertyKey=propertyValue
```
We should extract:
```
key1=value1
key2=value2
key3=value3
```

----

Related: in .NET we are _not_ ignoring properties, so we extract them as part of the value instead of separately:
```
key1=value1;property1;property2
key2=value2
key3=value3; propertyKey=propertyValue
```

This will be problematic because when we inject those values back into a baggage header, we will end up encoding the semi-colons:
```
baggage: key1=value1%3Bproperty1%3Bproperty2,key2=value2,key3=value3%3BpropertyKey=propertyValue
```
This will result in the properties being part of the values for any other downstream service instead of being separate metadata, which is... not great.

dmehala: addressed in 5b8f0b97b03091afdd7cb536e7a153e87001c25a

zacharycmontoya (approved):
LGTM! Thank you for engaging with all my feedback

dmehala:
@zacharycmontoya, thank you for reviewing. Your feedback really helped clarify the RFC and the expected behavior. I appreciate it.

PR #184, opened 2025-02-09
feat: store tracer configuration in an in-memory file

dmehala:
# Description

Resolves [APMAPI-1066]

[APMAPI-1066]: https://datadoghq.atlassian.net/browse/APMAPI-1066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/datadog/platform_util.cpp
+++ b/src/datadog/platform_util.cpp
@@ -1,6 +1,5 @@
 #include "platform_util.h"
 
-#include <cassert>
 // clang-format off
 #if defined(__x86_64__) || defined(_M_X64)
 #  define DD_SDK_CPU_ARCH "x86_64"
@@ -103,7 +102,7 @@ std::tuple<std::string, std::string> get_windows_info() {
   // application manifest, which is the lowest version supported by the
   // application. Use `RtlGetVersion` to obtain the accurate OS version
   // regardless of the manifest.
-  using RtlGetVersion = auto(*)(LPOSVERSIONINFOEXW)->NTSTATUS;
+  using RtlGetVersion = auto (*)(LPOSVERSIONINFOEXW)->NTSTATUS;
 
   RtlGetVersion func =
       (RtlGetVersion)GetProcAddress(GetModuleHandleA("ntdll"), "RtlGetVersion");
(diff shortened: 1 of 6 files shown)


src/datadog/platform_util.cpp
```diff
@@ -224,5 +228,58 @@ int at_fork_in_child(void (*on_fork)()) {
...
+
+  size_t written = write(fd, data.data(), data.size());
+  if (written != data.size()) return false;
+
+  return fcntl(fd, F_ADD_SEALS,
+               F_SEAL_SHRINK | F_SEAL_GROW | F_SEAL_WRITE | F_SEAL_SEAL) == 0;
+}
+
+Expected<InMemoryFile> InMemoryFile::make(StringView name) {
+  int fd = memfd_create(name.data(), MFD_CLOEXEC | MFD_ALLOW_SEALING);
```
pablomartinezbernardo, line 264: Question: The RFC suggest to use libdatadog for this. Is this temporary or will the cpp tracer now follow that advice?

src/datadog/tracer.cpp
```diff
@@ -110,6 +114,42 @@ std::string Tracer::config() const {
   return config.dump();
 }
 
+void Tracer::store_config() {
```
pablomartinezbernardo, line 117: Can this lead to incorrect usage? If store_config is called twice, will that result in multiple descriptors for this process? If so, what are the implications of that?

pablomartinezbernardo (approved):
LGTM, because although there's a couple of questions, being aware of those concerns is enough

src/datadog/platform_util.cpp
dmehala: It suggest, doesn't enforce. Since `dd-trace-cpp` has no dependency on `libdatadog`, it will not be that efficient to use it.

src/datadog/tracer.cpp
dmehala: Every call to this method will result in `N` file descriptors. Since `Tracer::store_config()` is a private method and we control how it’s called, I believe we can ensure it’s used correctly and avoid any unintended issues.

PR #185, opened 2025-02-11
feat: add telemetry logs

dmehala:
# Description
Add support telemetry logs

# Note to reviewers
This PR only adds the API for collecting and reporting logs. As of now, the library is not yet reporting logs because I am planning to refactor the whole telemetry module and error handling.


pablomartinezbernardo (approved):

include/datadog/telemetry/telemetry.h
```diff
@@ -16,19 +16,40 @@ class TracerTelemetry;
 
 namespace telemetry {
 
-class Telemetry {
+/// The telemetry class is responsible for handling internal telemetry data to
+/// track Datadog product usage. It _can_ collect and report logs and metrics.
+///
```
dubloom, line 21: Don't we have private headers ? If yes, is there a reason to put the file in the public headers and say do not use ?

dubloom (commented):
LGTM overall. Don't we want to add a test for logs payload as well ? Maybe you want to do it after the refactoring. I could get that.

PR #194, opened 2025-03-31
[part 2] refactor!(telemetry): move telemetry logic from `DatadogAgent`

dmehala:
## Description
This PR continues the work of refactoring the telemetry module to provide a clear and structured telemetry API. The goal is to make it easier for Datadog engineers to collect telemetry metrics and logs on the usage of our products.

Specifically, this PR moves the telemetry logic out of `DatadogAgent` and into a the telemetry module.
For reference, Part 1 of this refactoring can be found here: [#166](https://github.com/DataDog/dd-trace-cpp/pull/166).

### DIFF
--- a/src/datadog/tracer.cpp
+++ b/src/datadog/tracer.cpp
@@ -46,11 +46,11 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
                                     : RuntimeID::generate()),
       signature_{runtime_id_, config.defaults.service,
                  config.defaults.environment},
-      tracer_telemetry_(std::make_shared<TracerTelemetry>(
-          config.telemetry.enabled, config.clock, logger_, signature_,
-          config.integration_name, config.integration_version)),
-      config_manager_(
-          std::make_shared<ConfigManager>(config, tracer_telemetry_)),
+      telemetry_{std::make_shared<telemetry::Telemetry>(
+          config.telemetry, logger_, config.http_client,
+          std::vector<std::shared_ptr<telemetry::Metric>>{},
+          *config.event_scheduler, config.agent_url)},
+      config_manager_(std::make_shared<ConfigManager>(config, telemetry_)),
       collector_(/* see constructor body */),
       span_sampler_(
           std::make_shared<SpanSampler>(config.span_sampler, config.clock)),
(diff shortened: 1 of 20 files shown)


pablomartinezbernardo (approved):

src/datadog/tracer.cpp
```diff
@@ -74,14 +74,11 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
...
-        std::make_shared<DatadogAgent>(agent_config, tracer_telemetry_,
-                                       config.logger, signature_, rc_listeners);
+    auto agent = std::make_shared<DatadogAgent>(
+        agent_config, telemetry_, config.logger, signature_, rc_listeners);
     collector_ = agent;
 
-    if (tracer_telemetry_->enabled()) {
-      agent->send_app_started(config.metadata);
-    }
+    telemetry_->send_app_started(config.metadata);
```
dubloom, line 81: I understand removing the if to eval the condition directly in `send_telemetry` is less verbose, however I would see two disadvantages:
- With just a quick look at the code, you can think the telemetry is sent anyway
- It is less performant as we might do two functions call to not send the telemetry in the end. In the send_app_started it is negligible but if the telemetry is more used anywhere else it can have a (small) impact.

dubloom (commented):
This PR introduces a coupling between tracer_telemetry and telemetry which does not seem standard in comparison with the other tracers. Is there any objective or removing the tracer_telemetry later ? I would personally prefer if everything was centralized.

src/datadog/tracer.cpp
dubloom: In regards of the change of the 3rd part, this might not be relevant anymore (as the call will do nothing).

dmehala: when the telemetry is disable, every call will be noop. Also in the future, the tracer will not send the app-started event, instead, will add an event that the telemetry module will send whenever it feels is the right time.

dubloom: That's what I was thinking. Thank you for your confirmation. I resolve this comment.

src/datadog/tracer_telemetry.cpp
```diff
@@ -61,49 +61,24 @@ TracerTelemetry::TracerTelemetry(
...
+        internal_metrics,
     const std::vector<std::shared_ptr<telemetry::Metric>>& user_metrics)
     : enabled_(enabled),
       clock_(clock),
       logger_(logger),
       host_info_(get_host_info()),
       tracer_signature_(tracer_signature),
       integration_name_(integration_name),
-      integration_version_(integration_version),
-      user_metrics_(user_metrics) {
```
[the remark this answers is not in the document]
dmehala: That's a mistake, thank you for raising this. Addressed in b493ceea1a56809d6098f2ac2dc8e979652143e1

dmehala:
@dubloom the coupling in temporary. While we aim for standardization and feature parity, only public facing code and API are bound by that requirement.

PR #197, opened 2025-03-31
Correlate traces to profiles with the OTel ebpf profiler

elsakeirouz:
# Description
Based on Elastic's work to correlate traces to profiles for Java using the ebpf profiler.


elsakeirouz:
What do we think about using something like `#ifdef linux/unix` for parts that might not be relevant on Windows/MacOS? To be fair, since this code specifically is meant to work with the ebpf profiler on Linux, and will eventually only run if prompted, I do not see the point of compiling it on other OSs.

dubloom:
> What do we think of using something like `#ifdef linux/unix` for parts that might not be relevant on Windows/MacOS? To be fair, since this code specifically is meant to work with the ebpf profiler on Linux, and will eventually only run if prompted, I do not see the point of compiling it on other OSs.

My 2 cents here: if the code does not add a big compile time, I think we will pay more in code readibility by adding #ifdef (as the code is split into multiple files) than we will win in compile time. It can be worth adding some comments to tell it will only work in Linux though.
(Note this is my personal opinion and I'm not really a contributor of dd-trace-cpp atm).

elsakeirouz:
> > What do we think of using something like `#ifdef linux/unix` for parts that might not be relevant on Windows/MacOS? To be fair, since this code specifically is meant to work with the ebpf profiler on Linux, and will eventually only run if prompted, I do not see the point of compiling it on other OSs.
> 
> My 2 cents here: if the code does not add a big compile time, I think we will pay more in code readibility by adding #ifdef (as the code is split into multiple files) than we will win in compile time. It can be worth adding some comments to tell it will only work in Linux though. (Note this is my personal opinion and I'm not really a contributor of dd-trace-cpp atm).

Noted, thanks for your input!

elsakeirouz:
FYI, the global variables' naming is not set in stone. We plan on moving towards something more generic that does not use the elastic name.

Additionally, I am in the process of a writing a problem statement regarding the way we share process data since dd-trace-cpp already shares that data through memfds.

test/test_tracer.cpp
```diff
@@ -1566,6 +1566,32 @@ TEST_CASE("128-bit trace IDs") {
...
+  }
+
+  SECTION("is available when enabled") {
+    config.correlate_full_host_profiles = true;
+    auto finalized_config = finalize_config(config);
+    REQUIRE(finalized_config);
+    Tracer tracer{*finalized_config};
+    REQUIRE(elastic_apm_profiling_correlation_process_storage_v1 != nullptr);
+    auto span = tracer.create_span();
+    REQUIRE(elastic_apm_profiling_correlation_tls_v1 != nullptr);
```
elsakeirouz, line 1590: Could add more checks here for the actual stored values.

elsakeirouz:
What's left to do here is figure how we want to handle runtime-ids in case of forks.

test/test_tracer.cpp
```diff
@@ -1566,6 +1566,48 @@ TEST_CASE("128-bit trace IDs") {
...
+    Tracer tracer{*finalized_config};
+    REQUIRE(elastic_apm_profiling_correlation_process_storage_v1 == nullptr);
+  }
+
+  SECTION("is available when enabled") {
+    config.correlate_full_host_profiles = true;
+    auto finalized_config = finalize_config(config);
+    REQUIRE(finalized_config);
+    Tracer tracer{*finalized_config};
+    REQUIRE(elastic_apm_profiling_correlation_process_storage_v1 != nullptr);
```
elsakeirouz, line 1588: Since we have not yet defined what the process storage is going to look like, I am not validating the data that is stored in there. Once that is specified, we can look into decoding the contents.

cmake/compiler/clang.cmake
```diff
@@ -5,6 +5,26 @@
...
 
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  if (CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+    set(TLS_DIALECT desc)
+  elseif (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")
+    set(TLS_DIALECT gnu2)
+  else()
+    message(FATAL_ERROR "Only aarch64 and x86-64 are supported (found: ${CMAKE_SYSTEM_PROCESSOR})")
+  endif()
+
```
r1viollet, line 16: This part could be conditioned to an expected $TLS_DIALECT set above

src/datadog/span.cpp
```diff
@@ -40,6 +41,16 @@ Span::~Span() {
     data_->duration = now - data_->start;
   }
 
+#ifdef __linux__
+  // When a span is finished, we must update the span_id to its parent's.
+  if (elastic_apm_profiling_correlation_process_storage_v1 != nullptr &&
+      parent_id().has_value()) {
+    elastic_apm_profiling_correlation_tls_v1->valid = 0;
+    elastic_apm_profiling_correlation_tls_v1->span_id = parent_id().value();
+    elastic_apm_profiling_correlation_tls_v1->valid = 1;
```
ivoanjo, line 50: For future reference, if anyone's curious, I was reviewing this with @elsakeirouz and we noticed we're missing some synchronization here, [per the spec](https://github.com/elastic/apm/blob/bd5fa9c1/specs/agents/universal-profiling-integration.md#concurrency-safe-updates):

> Note that APM-agents must make sure that compilers do not reorder the steps listed above.

xlamorlette-datadog:
This PR was created on March 31, 2025 (18 months ago) and last updated on July 24, 2025 (14 months ago). I’m therefore closing it as obsolete.

ivoanjo:
Thanks @elsakeirouz for showing us the way here! This early work was super helpful in adding correlation with the eBPF Profiler with many languages beyond just dd-trace-cpp :D

PR #199, opened 2025-04-01
[part 4] fix(telemetry): report logs

dmehala:
# Description
Fix telemetry logs and add support for stacktrace reporting.

Changes:
 - Add unit tests.
 - Report `stack_trace` and `timestamp`.

### DIFF
--- a/include/datadog/telemetry/telemetry.h
+++ b/include/datadog/telemetry/telemetry.h
@@ -87,4 +87,10 @@ void report_warning_log(std::string message);
 /// @param message The error message.
 void report_error_log(std::string message);
 
+/// Report internal error message to Datadog.
+///
+/// @param message The error message.
+/// @param stacktrace Stacktrace leading to the error.
+void report_error_log(std::string message, std::string stacktrace);
+
 }  // namespace datadog::telemetry
--- a/src/datadog/telemetry/log.h
+++ b/src/datadog/telemetry/log.h
@@ -1,3 +1,6 @@
+#include <datadog/optional.h>
+#include <datadog/string_view.h>
+
 #include <string>
 
 namespace datadog::telemetry {
@@ -7,6 +10,20 @@ enum class LogLevel : char { ERROR, WARNING };
 struct LogMessage final {
   std::string message;
   LogLevel level;
+  tracing::Optional<std::string> stacktrace;
+  std::chrono::seconds::rep timestamp;
 };
+
+inline tracing::StringView to_string(LogLevel level) {
+  switch (level) {
+    case LogLevel::ERROR:
+      return "ERROR";
+    case LogLevel::WARNING:
+      return "WARNING";
+  }
+
+  // Unreachable.
+  return "";
+}
 
 }  // namespace datadog::telemetry
--- a/src/datadog/telemetry/telemetry.cpp
+++ b/src/datadog/telemetry/telemetry.cpp
@@ -116,4 +116,14 @@ void report_error_log(std::string message) {
              instance());
 }
 
+void report_error_log(std::string message, std::string stacktrace) {
+  std::visit(details::Overload{
+                 [&](Telemetry& telemetry) {
+                   telemetry.log_error(message, stacktrace);
+                 },
+                 [](auto&&) {},
+             },
+             instance());
+}
+
 }  // namespace datadog::telemetry
--- a/src/datadog/telemetry/telemetry_impl.cpp
+++ b/src/datadog/telemetry/telemetry_impl.cpp
@@ -201,6 +201,10 @@ void Telemetry::log_error(std::string message) {
   tracer_telemetry_->log(std::move(message), LogLevel::ERROR);
 }
 
+void Telemetry::log_error(std::string message, std::string stacktrace) {
+  tracer_telemetry_->log(std::move(message), LogLevel::ERROR, stacktrace);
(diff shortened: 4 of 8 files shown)


dubloom:
Hello there, I will review the 3 PRs on telemetry tomorrow morning !

pablomartinezbernardo (approved):

include/datadog/telemetry/telemetry.h
```diff
@@ -87,4 +87,10 @@ void report_warning_log(std::string message);
 /// @param message The error message.
 void report_error_log(std::string message);
```
dubloom, line 88:
```suggestion
/// Report internal error message to Datadog.
///
/// @param message The error message.
/// @param stacktrace Stacktrace leading to the error.
void report_error_log(std::string message, std::optional<std::string> stacktrace = std::nullopt);

```

src/datadog/telemetry/telemetry_impl.cpp
```diff
@@ -201,6 +201,10 @@ void Telemetry::log_error(std::string message) {
   tracer_telemetry_->log(std::move(message), LogLevel::ERROR);
 }
 
+void Telemetry::log_error(std::string message, std::string stacktrace) {
```
dubloom, line 204:
```suggestion
void Telemetry::log_error(std::string message, std::optional<std::string> stacktrace) {
```

src/datadog/telemetry/telemetry_impl.h
```diff
@@ -58,6 +58,7 @@ class Telemetry final {
   ///
   /// @param message The error message.
   void log_error(std::string message);
```
dubloom, line 60:
```suggestion
  void log_error(std::string message, std::optional<std::string> stacktrace = std::nullopt);
```

src/datadog/telemetry/telemetry.cpp
```diff
@@ -116,4 +116,14 @@ void report_error_log(std::string message) {
              instance());
 }
 
+void report_error_log(std::string message, std::string stacktrace) {
```
dubloom, line 119: If my other suggestions are accepted, here we could use:
```suggestion
void report_error_log(std::string message, std::optional<std::string> stacktrace) {
  std::visit(details::Overload{
                 [&](Telemetry& telemetry) { telemetry.log_error(message, stacktrace); },
                 [](auto&&) {},
             },
             instance());
}
```

dubloom (approved):
LGTM,  I made suggestion mainly about having one function with optional stacktrace everywhere but I'm still good if you don't want to do that.

src/datadog/telemetry/log.h
```diff
@@ -7,6 +10,20 @@ enum class LogLevel : char { ERROR, WARNING };
...
+inline tracing::StringView to_string(LogLevel level) {
+  switch (level) {
+    case LogLevel::ERROR:
+      return "ERROR";
+    case LogLevel::WARNING:
+      return "WARNING";
+  }
+
+  // Unreachable.
+  return "";
```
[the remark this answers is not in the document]
dmehala: That is intentional. `default` acts as a sink which can be error-prone when a new value to the enum is added. With this implementation, if a new value is added, the compiler error will be raised.

include/datadog/telemetry/telemetry.h
dmehala: The lib embrace overloading instead of default values. There are several reasons to prefer overloading to default arguments. Smarter people already wrote good arguments, [Google's C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Default_Arguments):

> Default arguments are allowed on non-virtual functions when the default is guaranteed to always have the same value. Follow the same restrictions as for [function overloading](https://google.github.io/styleguide/cppguide.html#Function_Overloading), and prefer overloaded functions if the readability gained with default arguments doesn't outweigh the downsides below.
>
> Often you have a function that uses default values, but occasionally you want to override the defaults. Default parameters allow an easy way to do this without having to define many functions for the rare exceptions. Compared to overloading the function, default arguments have a cleaner syntax, with less boilerplate and a clearer distinction between 'required' and 'optional' arguments.
>
> Defaulted arguments are another way to achieve the semantics of overloaded functions, so all the [reasons not to overload functions](https://google.github.io/styleguide/cppguide.html#Function_Overloading) apply.
>
> The defaults for arguments in a virtual function call are determined by the static type of the target object, and there's no guarantee that all overrides of a given function declare the same defaults.
>
> Default parameters are re-evaluated at each call site, which can bloat the generated code. Readers may also expect the default's value to be fixed at the declaration instead of varying at each call.
>
> Function pointers are confusing in the presence of default arguments, since the function signature often doesn't match the call signature. Adding function overloads avoids these problems.

I do agree some times I take shortcut, but that is usually located and with a plan to remove the technical debt.

src/datadog/telemetry/log.h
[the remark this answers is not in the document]
dubloom: That is true.

include/datadog/telemetry/telemetry.h
dubloom: Thank you for this context. My comment was more about understanding the choice of doing it this way and I thought that if the choice was made to use default parameter, at least the suggestions were already made. 
Thank you for your reply, i'll resolve the comment

PR #206, opened 2025-04-12
[part 8] refactor!(telemetry): rework metrics API

dmehala:
# Description
Redesigned the metrics API to eliminate the need for pre-registering a list of metrics.

Previously, the API relied on a pull-based approach where metrics had to be registered during the construction of the telemetry. The module would then periodically retrieve values from the registered metrics. While functional, this design was cumbersome - especially for metrics involving dynamic tags.

This refactoring introduces a push-based mechanism. Calls to `rate`, `counter`, or `distribution` now directly send data to the telemetry instance, removing the need for upfront registration.

This approach simplifies usage and improves flexibility.


include/datadog/telemetry/telemetry.h
```diff
@@ -78,4 +73,89 @@ void report_error_log(std::string message);
...
+/// Increments the specified counter by 1.
+///
+/// @param `counter` the counter to increment.
+void increment(const Counter& counter);
+
+/// Increments the specified counter by 1.
+///
+/// @param `counter` the counter to increment.
+/// @param `tags` the distribution tags.
+void increment(const Counter& counter, const std::vector<std::string>& tags);
```
r1viollet, line 91: minor: should we consider spans of string views ?

include/datadog/telemetry/telemetry.h
```diff
@@ -78,4 +73,89 @@ void report_error_log(std::string message);
...
+/// @param `counter` the counter to update.
+/// @param `tags` the distribution tags.
+/// @param `value` the value to assign to the counter.
+void set(const Counter& counter, const std::vector<std::string>& tags,
+         uint64_t value);
+
+}  // namespace counter
+
+/// The `rate` namespace provides support for rate metrics-values.
+/// Rates can be useful for tracking the total number of an event occurences  in
```
r1viollet, line 121: I had to re-google `Which is correct, occurence or occurrence?` :smile:

include/datadog/telemetry/metrics.h
```diff
@@ -1,146 +1,34 @@
...
- public:
-  CounterMetric(std::string name, std::string scope,
-                std::vector<std::string> tags, bool common);
-  void inc();
-  void add(uint64_t amount);
+namespace details {
+enum class MetricType : char { counter, rate, distribution };
+}
+
+/// TODO: pre-compute hash?
```
r1viollet, line 12: good question! Would it be worth it ?

src/datadog/common/hash.cpp
```diff
@@ -0,0 +1,317 @@
...
+#pragma warning(push)
+#pragma warning(disable : 4267)
+#endif
+
+constexpr bool ALLOW_UNALIGNED_READS = true;
+
+namespace datadog::common {
+namespace details {
+
+void SpookyHash::Short(const void *message, size_t length, uint64 *hash1,
```
r1viollet, line 15: Any reason to use spooky versus other hash implementations like [xxh3](https://github.com/Cyan4973/xxHash?tab=readme-ov-file) ? 
Feel free to discard the question if this is not so relevant.

dmehala: Hey @r1viollet , thanks for chiming in. I must admit I wasn't aware of xxhash. I favored SpookyHash mainly because the performance for long strings is slightly better than CityHash, and the append interface makes it less error-prone for combining hashes. I'll take a look at xxhash. Thanks!

include/datadog/telemetry/metrics.h
dmehala: Only data can answer that question. Until then, I guess I'll keep it simple.

include/datadog/telemetry/telemetry.h
dmehala: spans, like `std::spans`? If so, unfortunately we are stuck with C++17 at the moment.

r1viollet: Ah yes, we used to rely on [tcb spans](https://github.com/DataDog/ddprof/blob/78b6920991c720b7c9f2b21af07f71ab3ae8ada0/third_party/tcb/span.hpp) when we did not have c++ 20
Vectors seem fine in the mean time

include/datadog/telemetry/telemetry.h
r1viollet: just to clarify there is a minor typo

include/datadog/telemetry/metrics.h
r1viollet: how do you usually benchmark this ?

include/datadog/telemetry/telemetry.h
dmehala, line 121:
```suggestion
/// Rates can be useful for tracking the total number of an event occurrences  in
```

include/datadog/telemetry/telemetry.h
```diff
@@ -78,4 +73,89 @@ void report_error_log(std::string message);
 /// @param stacktrace Stacktrace leading to the error.
 void report_error_log(std::string message, std::string stacktrace);
 
+/// The `counter` namespace provides functions to track values.
+/// Counters can be useful for tracking the total number of an event occuring in
```
dmehala, line 77:
```suggestion
/// Counters can be useful for tracking the total number of an event occurring in
```

src/datadog/tracer.cpp
```diff
@@ -380,7 +381,8 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
   }
 
   const auto span_data_ptr = span_data.get();
-  telemetry::metrics().tracer.trace_segments_created_continued.inc();
+  telemetry::counter::increment(metrics::tracer::trace_segments_created,
+                                {"new_continued:new"});
```
pablomartinezbernardo, line 385:
```suggestion
                                {"new_continued:continued"});
```

src/datadog/tracer.cpp
```diff
@@ -441,9 +443,13 @@ Expected<void> Tracer::inject(const Baggage& baggage, DictWriter& writer) {
         err->with_prefix("failed to serialize all baggage items: "));
 
     if (err->code == Error::Code::BAGGAGE_MAXIMUM_BYTES_REACHED) {
-      telemetry::metrics().tracer.baggage_bytes_exceeded.inc();
+      telemetry::counter::increment(
+          metrics::tracer::context_header_truncated,
+          {"truncation_reason:baggage_item_count_exceeded"});
```
pablomartinezbernardo, line 448:
```suggestion
          {"truncation_reason:baggage_byte_count_exceeded"});
```

src/datadog/tracer.cpp
```diff
@@ -441,9 +443,13 @@ Expected<void> Tracer::inject(const Baggage& baggage, DictWriter& writer) {
...
     } else if (err->code == Error::Code::BAGGAGE_MAXIMUM_ITEMS_REACHED) {
-      telemetry::metrics().tracer.baggage_items_exceeded.inc();
+      telemetry::counter::increment(
+          metrics::tracer::context_header_truncated,
+          {"truncation_reason:baggage_byte_count_exceeded"});
```
pablomartinezbernardo, line 452:
```suggestion
          {"truncation_reason:baggage_item_count_exceeded"});
```

src/datadog/telemetry/telemetry_impl.h
```diff
@@ -9,16 +9,29 @@
...
 #include "json.hpp"
 #include "log.h"
+#include "metric_context.h"
 #include "platform_util.h"
 
 namespace datadog::telemetry {
 
+// This uses a reference_wrapper so references to internal metric values can
+// be captured, and be iterated trivially when the values need to be
+// snapshotted and published in telemetry messages.
```
dmehala, line 23:
```suggestion
```

pablomartinezbernardo (approved):

PR #208, opened 2025-04-21
perf: improve mutex contention

dmehala:
# Description

The telemetry refactoring exposed high mutex contention in our NGINX integration during shutdown/reload phase. This commit addresses mutex contention in the curl integration used by the default http client.


pablomartinezbernardo (approved):
Can't find any problematic scenario, very good job

PR #214, opened 2025-05-12
feat: support origin detection

### LINKED ISSUE #193: [FEATURE]: Support origin discovery via entity-id or container-id
### Describe the goal of the feature

Looking at dd-trace-python, I see that library performs container discovery by examining the `/proc/` filesystem, and populates the `Datadog-Entity-Id` and `Datadog-Container-Id` headers accordingly.

Perhaps discovering these is out of scope of a C implementation, but would not at least looking for environment variables, `DD_CONTAINER_ID` and/or `DD_ENTITY_ID` be helpful?  

(Getting container id is not guaranteed to work in cgroups2, depending on namespace, but the cgroup inode number apparently does...)

I also see the python lib populate the Datadog-External-Env with the `DD_EXTERNAL_ENV` environment variable, although I don't know what that is supposed to do...

### Is your feature request related to a problem?

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

dmehala:
# Description
Origin Detection allows to detect where the contaienr traces come from, and add container tags automatically to the local root span.

Resolves #193 

# Motivation
Feature request #193 

# Additional Notes

Jira ticket: [APMAPI-1404]

[APMAPI-1404]: https://datadoghq.atlassian.net/browse/APMAPI-1404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/datadog/platform_util.cpp
+++ b/src/datadog/platform_util.cpp
@@ -314,7 +314,8 @@ Optional<ino_t> get_inode(std::string_view path) {
 // whether the binary is running in host or not. However, it does not work when
 // running in a Docker in Docker environment.
 bool is_running_in_host_namespace() {
-  // linux procfs file that represents the cgroup namespace of the current process
+  // linux procfs file that represents the cgroup namespace of the current
+  // process.
   if (auto inode = get_inode("/proc/self/ns/cgroup")) {
     return *inode == HOST_CGROUP_NAMESPACE_INODE;
   }


src/datadog/datadog_agent_config.cpp
```diff
@@ -144,6 +145,12 @@ Expected<FinalizedDatadogAgentConfig> finalize_config(
   result.metadata[ConfigName::AGENT_URL] =
       ConfigMetadata(ConfigName::AGENT_URL, url, origin);
 
+  /// Starting Agent X, the admission controller inject a unique identifier
```
dmehala, line 148: - [ ] find the datadog agent version

src/datadog/platform_util.h
```diff
@@ -72,5 +72,33 @@ std::string get_process_name();
...
+///
+/// @param source The input from which to read the Docker container ID.
+/// @return An Optional containing the Docker container ID if found, otherwise
+/// nothing.
+Optional<std::string> find_docker_container_id(std::istream& source);
+
+/// Function to retrieve the container metadata.
+///
+/// @return A `ContainerMetadata` object containing metadata of the container in
+/// which the current process is running.
```
dubloom, line 98:
```suggestion
/// @return A `ContainerID` object containing id of the container in
/// which the current process is running.
```

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+  }
+
+  return buf.st_ino;
+}
+
+// Host namespace inode number are hardcoded, which allows for dectection of
+// whether the binary is running in host or not. However, it does not work when
+// running in a Docker in Docker environment.
+bool is_running_in_host_namespace() {
+  if (auto inode = get_inode("/proc/self/ns/cgroup")) {
```
dubloom, line 317:
```suggestion
bool is_running_in_host_namespace() {
  // linux procfs file that represents the cgroup namespace of the current process
  if (auto inode = get_inode("/proc/self/ns/cgroup")) {
```

dubloom: i'm ready to hear this is overkill but I think it can be useful for people less used to this kind of low level processing

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+  if (buf.f_type == CGROUP_SUPER_MAGIC)
+    return Cgroup::v1;
+  else if (buf.f_type == CGROUP2_SUPER_MAGIC)
+    return Cgroup::v2;
+
+  return nullopt;
+}
+
+Optional<std::string> find_docker_container_id_from_cgroup() {
+  constexpr std::string_view cgroup_path = "/proc/self/cgroup";
```
dubloom, line 340: Do we gain something by not directly writing the `"/proc/self/cgroup" in the ifstream parameters ? I'm totally OK with staying like that though.

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+
+  auto cgroup_fd = std::ifstream(cgroup_path.data(), std::ios::in);
+  if (!cgroup_fd.is_open()) return nullopt;
+
+  return find_docker_container_id(cgroup_fd);
+}
+#endif
+}  // namespace
+
+Optional<std::string> find_docker_container_id(std::istream& source) {
```
dubloom, line 350: It seems that this function is never called in a non unix env. Therefore, is there a reason why it is defined anyway ? Readability ?

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+
+  if (buf.f_type == CGROUP_SUPER_MAGIC)
+    return Cgroup::v1;
+  else if (buf.f_type == CGROUP2_SUPER_MAGIC)
+    return Cgroup::v2;
+
+  return nullopt;
+}
+
+Optional<std::string> find_docker_container_id_from_cgroup() {
```
dubloom, line 339:
```suggestion
Optional<std::string> find_docker_container_id_from_cgroup_v1() {
```

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+    return nullopt;
+  }
+
+  auto maybe_cgroup = get_cgroup_version();
+  if (!maybe_cgroup) return nullopt;
+
+  ContainerID id;
+  switch (*maybe_cgroup) {
+    case Cgroup::v1: {
+      if (auto maybe_id = find_docker_container_id_from_cgroup()) {
```
dubloom, line 385:
```suggestion
      if (auto maybe_id = find_docker_container_id_from_cgroup_v1()) {
```

src/datadog/datadog_agent.cpp
```diff
@@ -156,10 +157,36 @@ DatadogAgent::DatadogAgent(
...
+  headers_.emplace("Datadog-Meta-Tracer-Version",
+                   tracer_signature.library_version);
+
+  // Origin Detection headers are not necessary when Unix Domain Socket (UDS)
+  // is used to communicate with the Datadog Agent.
+  if (!contains(config.url.scheme, "unix")) {
+    if (auto container_id = container::get_id()) {
+      if (container_id->type == container::ContainerID::Type::container_id) {
+        headers_.emplace("Datadog-Container-ID", container_id->value);
+        headers_.emplace("Datadog-Entity-Id", "ci-" + container_id->value);
```
dubloom, line 177:
```suggestion
        headers_.emplace("Datadog-Entity-Id", "cid-" + container_id->value);
```

According to [JIRA spec](https://datadoghq.atlassian.net/wiki/spaces/CONTP/pages/3731783774/New+Origin+Detection+Spec+for+Traces+and+DogStatsD)

dubloom (changes requested):
Most of my comments are nit/picky ones. 
However there is one env variable that is not set according to the specifications.

src/datadog/platform_util.cpp
dmehala: not really, it's a relic of an old implementation.

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+  else if (buf.f_type == CGROUP2_SUPER_MAGIC)
+    return Cgroup::v2;
+
+  return nullopt;
+}
+
+Optional<std::string> find_docker_container_id_from_cgroup() {
+  constexpr std::string_view cgroup_path = "/proc/self/cgroup";
+
+  auto cgroup_fd = std::ifstream(cgroup_path.data(), std::ios::in);
```
dmehala, line 342:
```suggestion
  auto cgroup_fd = std::ifstream("/proc/self/cgroup", std::ios::in);
```

src/datadog/platform_util.cpp
dmehala: This function is not cgroupv1 specific.

src/datadog/platform_util.cpp
dmehala: this function only exists for testing purposes, i'd like to avoid `#ifdef/#endif` in tests.

src/datadog/datadog_agent.cpp
dmehala: TBH, the spec is confusing. It mentions `Datadog-Entity-Id=ci-<CONTAINER_ID>` but for **backward compatibility** Agent would still need to be able to resolve `Datadog-Entity-Id=cid-<CONTAINER_ID>`, then proceed to talk about backward compatibility depending on `DD_EXTERNAL_ENV` existence _BUT_ that section is in red which means according to the color coding:
> Red solutions are solutions that have been refused

All in all, my understanding is th expected value is `Datadog-Entity-Id=ci-<CONTAINER_ID>` but for backward compatibility reason with the Agent(?) it must also/or support `Datadog-Entity-Id=cid-<CONTAINER_ID>`. @wdhif can you weight in?

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,127 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+    return Cgroup::v1;
+  else if (buf.f_type == CGROUP2_SUPER_MAGIC)
+    return Cgroup::v2;
+
+  return nullopt;
+}
+
+Optional<std::string> find_docker_container_id_from_cgroup() {
+  constexpr std::string_view cgroup_path = "/proc/self/cgroup";
+
```
dmehala, line 342:
```suggestion
```

src/datadog/datadog_agent.cpp
dmehala: Thanks @wdhif

src/datadog/platform_util.cpp
```diff
@@ -281,5 +286,126 @@ Expected<InMemoryFile> InMemoryFile::make(StringView) {
...
+Optional<std::string> find_docker_container_id_from_cgroup() {
+  auto cgroup_fd = std::ifstream("/proc/self/cgroup", std::ios::in);
+  if (!cgroup_fd.is_open()) return nullopt;
+
+  return find_docker_container_id(cgroup_fd);
+}
+#endif
+}  // namespace
+
+Optional<std::string> find_docker_container_id(std::istream& source) {
```
dubloom, line 350:
```suggestion
// this function only exists for testing purposes
Optional<std::string> find_docker_container_id(std::istream& source) {
```

dubloom (approved):
LGTM

src/datadog/platform_util.cpp
dmehala: That's litterally part of the header file: https://github.com/DataDog/dd-trace-cpp/blob/29f6a506f924f3058966158495b1812a3769c5c2/src/datadog/platform_util.h#L88

PR #218, opened 2025-06-06
fix(telemetry): send app-started asynchronously

dmehala:
# Description
This addresses a regression introduced in #205 that caused a noticeable delay during the first initialization of the library.


pablomartinezbernardo (approved):
LGTM

PR #227, opened 2025-08-05
build: do not `FetchContent nlohmann_json` if already found by `find_package`

sjanel:
# Description
This PR makes it possible to use `dd-trace-cpp` as a dependency when nlohmann_json has already been fetched content by a project.

It also renames `json` into `nlohmann_json` as it is more standard in other projects.

# Motivation

# Additional Notes
Bump version of `nlohmann_json` to 3.12.0.

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/cmake/deps/json.cmake
+++ b/cmake/deps/json.cmake
@@ -3,8 +3,10 @@ include(FetchContent)
 set(JSON_BuildTests OFF)
 set(JSON_Install OFF)
 
-FetchContent_Declare(json
-  URL https://github.com/nlohmann/json/releases/download/v3.11.2/json.tar.xz
+FetchContent_Declare(nlohmann_json
+  URL https://github.com/nlohmann/json/archive/refs/tags/v3.12.0.tar.gz
+  URL_HASH SHA256=4b92eb0c06d10683f7447ce9406cb97cd4b453be18d7279320f7b2f025c10187
+  FIND_PACKAGE_ARGS NAMES nlohmann_json
 )
 
-FetchContent_MakeAvailable(json)
+FetchContent_MakeAvailable(nlohmann_json)


sjanel:
> Thanks for your contribution @sjanel. While I'm not opposed to avoiding the download of nlohmann-json if it's already installed on the system, we should ensure that the installed version matches the required one to prevent potential compilation errors.

To ensure a minimum version is found, we can do something like:

```
find_package(nlohmann_json 3.11.2 QUIET)
```

cmake/deps/json.cmake
```diff
@@ -1,10 +1,14 @@
...
+  set(JSON_BuildTests OFF)
+  set(JSON_Install OFF)
 
-FetchContent_Declare(json
-  URL https://github.com/nlohmann/json/releases/download/v3.11.2/json.tar.xz
-)
+  FetchContent_Declare(nlohmann_json
+    URL https://github.com/nlohmann/json/archive/refs/tags/v3.12.0.tar.gz
+    URL_HASH SHA256=4b92eb0c06d10683f7447ce9406cb97cd4b453be18d7279320f7b2f025c10187
+  )
```
dmehala, line 11: Alright, I did [my homework](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#fetchcontent-and-find-package-integration). I guess a simpler solution (and more idiomatic) solution is:

```suggestion
  FetchContent_Declare(nlohmann_json
    URL https://github.com/nlohmann/json/archive/refs/tags/v3.12.0.tar.gz
    URL_HASH SHA256=4b92eb0c06d10683f7447ce9406cb97cd4b453be18d7279320f7b2f025c10187
    FIND_PACKAGE_ARGS NAMES nlohmann_json
  )
```

Adding `FIND_PACKAGE_ARGS` will help `FetchContent_*` to satisfy the dependency o with `find_package()`. If that dependency is not found, then it is built from source.

Can you have a try?

sjanel: Yep, I see that you require cmake >= 3.24, so `FIND_PACKAGE_ARGS` will be available.

cmake/deps/json.cmake
```diff
@@ -1,10 +1,15 @@
 include(FetchContent)
 
-set(JSON_BuildTests OFF)
-set(JSON_Install OFF)
+find_package(nlohmann_json CONFIG QUIET)
+if (NOT nlohmann_json_FOUND)
```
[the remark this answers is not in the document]
sjanel: done

dmehala (approved):
LGTM! Thank you for your contribution 🙇

dmehala:
Slight details, @sjanel can you sign your commit?

PR #233, opened 2025-08-15
fix: edge cases parsing of `DD_TAGS`

dmehala:
# Description
This pull request addresses a bug in the tag parsing logic where space-separated values were incorrectly considered even when comma separators were present in the `DD_TAGS` environment variable. According to the RFC:

"This is a (mostly) backwards-compatible change to DD_TAGS parsing where we use comma-separated tags if there are any commas present in the DD_TAGS variable, and use space-separation otherwise."

The fix ensures that the presence of commas takes precedence, aligning the parsing behavior with the intended specification.

## Motivation

The bug was identified during system tests, which rely on correct tag parsing to function as expected. This fix brings the implementation into compliance with the RFC and prevents inconsistent behavior in environments where mixed separators may be present.

### DIFF
--- a/src/datadog/parse_util.cpp
+++ b/src/datadog/parse_util.cpp
@@ -141,11 +141,10 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
 
     if (separator == std::string::npos) {
       key = std::string{trim(token)};
+      if (key.empty()) continue;
     } else {
       key = std::string{trim(token.substr(0, separator))};
-      if (key.empty()) {
-        continue;
-      }
+      if (key.empty()) continue;
       value = std::string{trim(token.substr(separator + 1))};
     }
 
@@ -183,7 +182,6 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
       break;
     } else if (input[i] == ' ') {
       separator = ' ';
-      break;
     }
   }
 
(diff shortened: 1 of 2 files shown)


zacharycmontoya (approved):
LGTM. Thanks for your patience 🙏🏼

src/datadog/parse_util.cpp
```diff
@@ -183,15 +182,14 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
-      break;
     }
   }
 
   if (separator == 0) {
     goto capture_all;
   }
 
-  for (; i < end; ++i) {
+  for (i = beg; i < end; ++i) {
```
dubloom, line 192: Can we put i = 0 instead ? I'm almost sure that beg is 0 at this time and it makes the code a slightly more readable

src/datadog/parse_util.cpp
```diff
@@ -141,11 +141,10 @@ Expected<std::unordered_map<std::string, std::string>> parse_tags(
...
     if (separator == std::string::npos) {
       key = std::string{trim(token)};
+      if (key.empty()) continue;
     } else {
       key = std::string{trim(token.substr(0, separator))};
-      if (key.empty()) {
-        continue;
-      }
+      if (key.empty()) continue;
       value = std::string{trim(token.substr(separator + 1))};
```
dubloom, line 148: Could it be interesting to ignore the key with no value ?

dubloom (approved):
Nits but LGTM

src/datadog/parse_util.cpp
dmehala: Make sens. Unfortunately, this PR has been opened for too long, and revisiting it for a nit is not a priority. However, please don’t hesitate to open a new PR if you'd like to revisit or continue the work

PR #234, opened 2025-08-16
fix(rc): rework remote configuration handling

dmehala:
This commit refactors the way remote configurations are applied to the tracer. Previously, if a remote configuration contained both valid and invalid fields, the valid parts could be partially applied, potentially leading to inconsistent behavior.

Now, the configuration is only applied if it is fully valid. Any invalid configuration is entirely rejected. This constitutes a breaking change.

This also report errors when an invalid configuration is rejected.

Changes:
  - Add fuzz test and input corpus for Remote Configuration.
  - Fix a crash when telemetry methods are invoked before initializing the module.
  - Improve unit test coverage of remote configuration.


test/test_config_manager.cpp
```diff
@@ -169,23 +402,154 @@ CONFIG_MANAGER_TEST("remote configuration handling") {
...
+          },
+          {
+              __LINE__,
+              "invalid value for `sample_rate` field",
+              R"("tracing_sampling_rules": [{"service": "bar", "resource": "yo", "sample_rate": "0.5", "provenance": "customer"}])",
+          },
+          {
+              __LINE__,
+              "invalid value for `tags` field",
+              R"("tracing_sampling_rules": [{"service": "bar", "resource": "yo", "sample_rate": "0.5", "provenance": "customer", "tags": "tag1"}])",
```
zacharycmontoya, line 489:
```suggestion
              R"("tracing_sampling_rules": [{"service": "bar", "resource": "yo", "sample_rate": 0.2, "provenance": "customer", "tags": "tag1"}])",
```

test/test_config_manager.cpp
```diff
@@ -169,23 +402,154 @@ CONFIG_MANAGER_TEST("remote configuration handling") {
...
+          },
+          {
+              __LINE__,
+              "invalid second rules",
+              R"("tracing_sampling_rules": [{"service": "bar", "resource": "yo", "sample_rate": "0.5", "provenance": "customer"}, {"foo": "bar"}])",
```
zacharycmontoya, line 494:
```suggestion
              R"("tracing_sampling_rules": [{"service": "bar", "resource": "yo", "sample_rate": 0.2, "provenance": "customer"}, {"foo": "bar"}])",
```

dmehala:
@dubloom I doubt the segmentation fault observed in Istio are related. It's worth trying tho.

src/datadog/config_manager.cpp
```diff
@@ -24,6 +24,15 @@ nlohmann::json to_json(const SpanDefaults& defaults) {
   return result;
 }
 
+nlohmann::json to_json(const std::vector<TraceSamplerRule>& rules) {
+  std::vector<nlohmann::json> j;
```
ygree, line 28: Maybe use `j.reserve(rules.size());` to avoid possible reallocations?

PR #236, opened 2025-09-04
fix(rc): handle null values in RC APM Tracing payloads

dmehala:
# Description
Ensures system-tests pass by properly ignoring null values APM Tracing RC payload.


dubloom (approved):
LGTM

PR #178, opened 2025-01-24
perf: parse w3c traceparent using a customer parser

dmehala:
# Description
`std::regex` is [notoriously known to be slow](https://github.com/HFTrader/regex-performance?tab=readme-ov-file#results). This custom parser implementation offers significantly better performance.

## Benchmark
Bench: https://quick-bench.com/q/XJw86Zh_SBDpQ3xhI2nrY-k-Xbc
Result:
image

## TODO
- [x] fuzz me!

PR #241, opened 2025-09-10
fix: reject traceparent headers with unsupported versions

dmehala:
# Description
This addresses a regression introduced in #178, where traceparent headers containing unsupported characters were not properly rejected and were incorrectly treated as valid.

[APMAPI-1599]

[APMAPI-1599]: https://datadoghq.atlassian.net/browse/APMAPI-1599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/test_tracer.cpp
+++ b/test/test_tracer.cpp
@@ -755,6 +755,19 @@ TEST_TRACER("span extraction") {
         {__LINE__, "invalid: non hex trace tag ID",
          "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-xy", // traceparent
          "malformed_traceflags"}, // expected_error_tag_value
+
+        {
+            __LINE__,
+            "invalid: non supported character in trace version 1/x",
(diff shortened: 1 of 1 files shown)


test/test_tracer.cpp
```diff
@@ -1191,6 +1191,31 @@ TEST_TRACER("span extraction") {
             nullopt,
             "0000000000000000",  // expected_datadog_w3c_parent_id,
         },
+
+        {
+            __LINE__,
+            "malformed traceparent 1/x",
```
zacharycmontoya, line 1197: It seems like this test would fit better in a previous test case, where you test error cases and assert that the expected_error_tag_value is `malformed_traceparent`. However this input would result in `invalid_version`

dmehala: Good catch, thank you @zacharycmontoya . Addressed in 071b25b60735bfdbf83b4cb1a63cc669e4e7c087

PR #281, opened 2026-02-19
chore!: add supported configurations

BridgeAR:
The configuration macro is changed to a registry that includes types and default values next to the configuration names.

This is used for compile time generation instead of using runtime checks for the configurations (environment variables).

For safety, failing to parse configurations is now going to log a warning instead. This aligns with other platforms / tracers to not crash a customer in production where their staging configuration might be valid and their production one not.

The macro must now be used for any new configuration added and a CI job verifies that all entries do exist. If a new env is added otherwise it should fail.

The CI will also verify that the generated file is up to date and matches the remote registry. It would fail, if either is not the case.

The behavior change is documented in the README.md.

TODO: Add Jira ticket

### DIFF
--- a/bin/supported-configurations.cpp
+++ b/bin/supported-configurations.cpp
@@ -1,0 +1,648 @@
+#include <algorithm>
+#include <array>
+#include <cctype>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <map>
+#include <optional>
+#include <regex>
+#include <set>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "json.hpp"
+
+namespace fs = std::filesystem;
+using nlohmann::json;
+
+namespace {
+
+const std::vector<std::string> kScanDirs = {"include", "src", "test", "examples",
+                                            "fuzz"};
+const std::set<std::string> kScanSuffixes = {".h", ".hh", ".hpp", ".c",
+                                             ".cc", ".cpp", ".cxx"};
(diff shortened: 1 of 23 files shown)


src/datadog/datadog_agent_config.cpp
```diff
@@ -51,7 +60,8 @@ Expected<DatadogAgentConfig> load_datadog_agent_env_config() {
 Expected<FinalizedDatadogAgentConfig> finalize_config(
     const DatadogAgentConfig& user_config,
     const std::shared_ptr<Logger>& logger, const Clock& clock) {
-  Expected<DatadogAgentConfig> env_config = load_datadog_agent_env_config();
+  Expected<DatadogAgentConfig> env_config =
+      load_datadog_agent_env_config(*logger);
   if (auto error = env_config.if_error()) {
     return *error;
   }
```
anna-git: this could go as well if no errors will be returned but logging instead

anna-git (commented):
Nice work, left a few comments. Mostly wondering if we should carry over a logger and instead return an enriched result type with potential errors to log along the side.

anna-git (approved):
thanks for addressing the comments!

bin/supported-configurations.cpp
```diff
@@ -0,0 +1,648 @@
...
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "json.hpp"
+
+namespace fs = std::filesystem;
+using nlohmann::json;
+
+namespace {
```
dmehala: useless anonymous namespace.

PR #294, opened 2026-03-20
feat: implement stable configuration

bm1549:
# Description

Phase 1 Stable Configuration support for dd-trace-cpp, rebased onto the `ConfigProvider` refactor from #321. Scalar `apm_configuration_default` keys only; Linux-only.

At tracer init, the tracer reads two YAML files:
- **Local** (user-managed): `/etc/datadog-agent/application_monitoring.yaml`
- **Fleet** (Agent-managed): `/etc/datadog-agent/managed/datadog-agent/stable/application_monitoring.yaml`

Each file may contain `apm_configuration_default`, a flat map of `DD_*` environment-variable-style keys to scalar values. These participate in configuration precedence: `fleet_stable > env > user/code > local_stable > default`. Fleet files may include a `config_id` reported in telemetry against fleet-sourced entries.

## How it's wired

This change plugs stable configuration into the `ConfigProvider` infrastructure landed in #321. Two `ConfigSource` implementations adapt parsed YAML into the provider's source list:

- `LocalStableConfigSource` (origin: `LOCAL_STABLE_CONFIG`)
- `FleetStableConfigSource` (origin: `FLEET_STABLE_CONFIG`, optional `config_id`)

`finalize_config` constructs both sources from the parsed YAML and passes them to `ConfigProvider` alongside `EnvironmentSource`. Every typed accessor on the provider (`get_string`, `get_bool`, `get_uint64`, `get_tags`, `get_propagation_styles*`) walks all three sources.

Keys covered: `DD_SERVICE`, `DD_ENV`, `DD_VERSION`, `DD_TAGS`, `DD_TRACE_ENABLED`, `DD_TRACE_STARTUP_LOGS`, `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`, `DD_APM_TRACING_ENABLED`, `DD_TRACE_BAGGAGE_MAX_ITEMS`, `DD_TRACE_BAGGAGE_MAX_BYTES`, `DD_TRACE_PROPAGATION_STYLE*`, `DD_TRACE_RESOURCE_RENAMING_*`.

## Out of scope (deferred to follow-ups)

- **Nested finalize_config integration**: `trace_sampler_config`, `span_sampler_config`, and `datadog_agent_config` finalize independently of the top-level `ConfigProvider`. Wiring stable config into them requires threading the sources through. Keys affected include `DD_TRACE_SAMPLE_RATE`, `DD_TRACE_RATE_LIMIT`, `DD_TRACE_SAMPLING_RULES`, `DD_SPAN_SAMPLING_RULES`, `DD_TRACE_AGENT_URL`. dd-trace-go's Phase 1 made the same choice; dd-trace-java covers them because its provider resolves universally.
- **Alias-precedence cleanup**: `get_propagation_styles_with_aliases` picks the first alias any source has and resolves only that alias. A specific alias set in env can shadow a more general alias set by fleet. Tracked separately; predates this PR (came in via #321).
- **Windows paths**: no peer reference for the `HKLM\SOFTWARE\Datadog\Datadog Agent` registry lookup in other native-parsing tracers. Defer to a follow-up aligned with the Agent's registry layout.
- **`apm_configuration_rules`** (targeting rules with selectors): not in any Phase 1 native implementation.

## Key design choices

- **YAML parsing via yaml-cpp 0.8.0** (`src/datadog/yaml_parser.{h,cpp}`). Linked PRIVATE in CMake; in Bazel, lives in `deps` with a comment documenting the asymmetry.
- **256KB file size limit** on each YAML file. Larger files are skipped with an error log.
- **YAML alias-bomb mitigation**: documents with more than 100 alias references are rejected before invoking `YAML::Load`. Heuristic guard, not a parser.
- **Non-scalar conversion safety**: `Node::as<std::string>()` calls are gated on `IsScalar()` and wrapped in try/catch.
- **Test-accessible `load_one`** in `stable_config.h` lets `TempDir`-based unit tests exercise file I/O paths directly.

## Files

| File | Change |
|------|--------|
| `cmake/deps/yaml.cmake` | New: yaml-cpp 0.8.0 FetchContent dependency |
| `BUILD.bazel`, `MODULE.bazel`, `WORKSPACE` | New: yaml-cpp Bazel dep |
| `CMakeLists.txt`, `test/CMakeLists.txt` | yaml-cpp + new test files |
| `src/datadog/yaml_parser.{h,cpp}` | New: yaml-cpp wrapper with alias-bomb and non-scalar guards |
| `src/datadog/stable_config.{h,cpp}` | New: `StableConfig`, Linux file paths, 256KB file loader |
| `src/datadog/stable_config_source.h` | New: `LocalStableConfigSource` and `FleetStableConfigSource` adapters |
| `src/datadog/tracer_config.cpp` | Wire the two stable-config sources into `ConfigProvider` |
| `src/datadog/telemetry/telemetry_impl.cpp` | Emit `config_id` per fleet-origin metadata entry |
| `test/test_yaml_parser.cpp` | YAML parser tests including non-scalar and alias-bomb cases |
| `test/test_stable_config.cpp` | StableConfig, file I/O via `TempDir`, source-adapter smoke tests |
| `test/system-tests/request_handler.{cpp,h}` | `/config` endpoint reads from `tracer_.config()` |

## Rebased onto #321

The earlier 40-commit history collapsed to one feature commit on top of `bm1549/config-provider`. Behavior matches the original PR's intent for in-scope keys; resolution machinery now lives in `ConfigProvider` and `ConfigSource` rather than in a 5-parameter `resolve_and_record_config` template. The legacy template was retired in #321.

## Testing

199,670 assertions / 172 test cases pass locally. Coverage includes:
- YAML parser happy paths, non-scalar `config_id` / map key handling, alias-bomb rejection
- `StableConfig::lookup`
- File I/O: valid YAML, 256KB rejection, malformed YAML on disk, missing file, unreadable file (POSIX-guarded)
- `LocalStableConfigSource` / `FleetStableConfigSource` adapter behavior including `config_id` forwarding
- `ConfigProvider` precedence chain (covered in #321's tests)

Related system-tests PR: DataDog/system-tests#6554.

# Motivation

Enable fleet-managed and user-managed APM configuration for C++ tracers (dd-trace-cpp, nginx-datadog, httpd-datadog). C++ is the last major language without stable config support. This is Phase 1; extended configs (sampler integration), nested finalizer wiring, and Windows support follow.

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -87,6 +87,11 @@ cc_library(
         "src/datadog/trace_segment.cpp",
         "src/datadog/trace_source.cpp",
         "src/datadog/tracer.cpp",
+        "src/datadog/stable_config.cpp",
+        "src/datadog/stable_config.h",
+        "src/datadog/stable_config_source.h",
+        "src/datadog/yaml_parser.cpp",
+        "src/datadog/yaml_parser.h",
         "src/datadog/tracer_config.cpp",
         "src/datadog/version.cpp",
         "src/datadog/w3c_propagation.cpp",
@@ -159,5 +164,10 @@ cc_library(
     deps = [
         "@com_google_absl//absl/strings",
         "@com_google_absl//absl/types:optional",
+        # yaml-cpp is an implementation detail; CMake links it PRIVATE.
+        # rules_cc's `implementation_deps` would be the analog, but the
+        # version pinned in MODULE.bazel does not yet support it, so we
+        # keep yaml-cpp in `deps` and document the asymmetry here.
+        "@yaml_cpp//:yaml-cpp",
     ],
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,10 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   include(cmake/compiler/gcc.cmake)
 endif ()
 
+# yaml-cpp must be included AFTER the compiler setup above, because
+# clang.cmake sets -stdlib=libc++ which yaml-cpp needs to inherit.
+include(cmake/deps/yaml.cmake)
+
 if (DD_TRACE_BUILD_FUZZERS)
   add_subdirectory(fuzz)
 endif ()
@@ -212,6 +216,8 @@ target_sources(dd-trace-cpp-objects
     src/datadog/tags.cpp
     src/datadog/tag_propagation.cpp
     src/datadog/threaded_event_scheduler.cpp
+    src/datadog/stable_config.cpp
+    src/datadog/yaml_parser.cpp
     src/datadog/tracer_config.cpp
     src/datadog/tracer.cpp
     src/datadog/trace_id.cpp
@@ -244,6 +250,7 @@ target_link_libraries(dd-trace-cpp-objects
     Threads::Threads
   PRIVATE
     dd-trace-cpp::specs
+    $<BUILD_INTERFACE:yaml-cpp>
 )
 
 set_target_properties(dd-trace-cpp-objects
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,3 +7,4 @@ bazel_dep(name = "abseil-cpp", version = "20260107.1", repo_name = "com_google_a
 bazel_dep(name = "bazel_skylib", version = "1.9.0")
 bazel_dep(name = "platforms", version = "1.1.0")
 bazel_dep(name = "rules_cc", version = "0.2.18")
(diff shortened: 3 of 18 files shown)


dmehala (changes requested):
I took a quick look, and this PR is a hard no. There are many problems, with the most obvious being the custom YAML parser. I *sincerely* appreciate your effort to add feature to the tracer. Coding agents can be useful, but they need an experienced person to guide them properly. Without that, the result is just a pile of code that is hard to maintain and fix.

PR #295, opened 2026-03-20
Implement stable session identifier headers for telemetry

khanayan123:
## Summary
Implements the Stable Service Instance Identifier RFC for the C++ SDK.

- Adds `DD-Session-ID` (= `runtime_id`) and conditional `DD-Root-Session-ID` headers to all telemetry requests (app-started, heartbeats, app-closing, etc.)
- `DD-Root-Session-ID` is only sent when it differs from `DD-Session-ID` (i.e., in child processes)
- Root session ID is stored in a Meyer's singleton (`static const` local), set once by the first `Tracer` to initialize. Immutable after init, so fork-safe.
- Fork propagation is automatic — the singleton value persists in child process memory

### Integration testing note
System-tests for `cpp_nginx` and `cpp_httpd` cannot validate session headers until the downstream modules (nginx-datadog, httpd-datadog) release a version that includes these dd-trace-cpp changes. The system-tests manifest entries remain `missing_feature` until then. Unit tests in this PR validate the implementation directly.

Companion system-tests PR: https://github.com/DataDog/system-tests/pull/6510


zacharycmontoya (commented):
Overall this change looks good. Before approving I'd like to get an opinion from another C++ dev on the thread-safety of the environment variable accesses. Once that's resolved, I think this PR is good to go

include/datadog/environment.h
```diff
@@ -110,6 +111,10 @@ StringView name(Variable variable);
 // `nullopt` if that variable is not set in the environment.
 Optional<StringView> lookup(Variable variable);
 
+// Set the specified environment `variable` to `value`. Does not overwrite if
+// already set (equivalent to setenv(..., 0) on POSIX).
+void set(Variable variable, StringView value);
```
xlamorlette-datadog: I think we could rename this function `set_if_not_set()`.

xlamorlette-datadog (commented):
Review on-going…

include/datadog/environment.h
```diff
@@ -81,7 +81,8 @@ namespace environment {
   MACRO(DD_APM_TRACING_ENABLED, BOOLEAN, true)                                 \
   MACRO(DD_TRACE_RESOURCE_RENAMING_ENABLED, BOOLEAN, false)                    \
   MACRO(DD_TRACE_RESOURCE_RENAMING_ALWAYS_SIMPLIFIED_ENDPOINT, BOOLEAN, false) \
-  MACRO(DD_EXTERNAL_ENV, STRING, "")
+  MACRO(DD_EXTERNAL_ENV, STRING, "")                                           \
+  MACRO(_DD_ROOT_CPP_SESSION_ID, STRING, "")
```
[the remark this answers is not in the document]
xlamorlette-datadog: Following internal discussion on Slack, I suggest adding, at the top of the file, a link to https://feature-parity.us1.prod.dog/configurations?viewType=configurations.

include/datadog/tracer_signature.h
```diff
@@ -33,15 +33,18 @@ namespace tracing {
...
+  std::string root_session_id;
   std::string default_service;
   std::string default_environment;
   std::string library_version;
   StringView library_language;
   StringView library_language_version;
 
   TracerSignature() = delete;
-  TracerSignature(RuntimeID id, std::string service, std::string environment)
+  TracerSignature(RuntimeID id, std::string root_session, std::string service,
```
xlamorlette-datadog: Beware: you are changing the constructor interface.
So you need to change the call in fuzz/remote-configuration/main.cpp.
Would it make sense to keep a constructor with only three parameters?
This would simplify many calls (for example the call from `init()` in `src/datadog/telemetry/telemetry.cpp`).

xlamorlette-datadog (commented):
Review still on-going…

include/datadog/environment.h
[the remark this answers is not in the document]
khanayan123: Sounds good

src/datadog/telemetry/telemetry_impl.cpp
```diff
@@ -60,6 +60,14 @@ HTTPClient::URL make_telemetry_endpoint(HTTPClient::URL url) {
   return url;
 }
 
+void set_session_headers(DictWriter& headers,
+                         const tracing::TracerSignature& sig) {
+  headers.set("DD-Session-ID", sig.runtime_id.string());
```
xlamorlette-datadog: _nit, opt_: store `sig.runtime_id.string()` in a variable to make intent clearer:
```
const std::string& session_id = sig.runtime_id.string();
headers.set("DD-Session-ID", session_id);
if (sig.root_session_id != session_id) {
  headers.set("DD-Root-Session-ID", sig.root_session_id);
}
```

src/datadog/telemetry/telemetry_impl.cpp
```diff
@@ -302,13 +310,15 @@ void Telemetry::app_started() {
   auto payload = app_started_payload();
 
   auto on_headers = [payload_size = payload.size(),
```
xlamorlette-datadog, line 320: The code lines 312-356 is very similar to the one lines 368-409.
The `on_response` handlers differ, and more counters are incremented in `send_payload()`, but I think that all what is in `send_payload()` is a more  complete version of what is in `app_started()`.
So I think that `app_started()` could be hugely simplified by simply calling `send_payload()` as follows:
```
void Telemetry::app_started() {
  send_payload("app-started", app_started_payload());
}
```

src/datadog/environment.cpp
```diff
@@ -11,19 +11,28 @@ namespace environment {
...
-  const char *value = std::getenv(name);
+  const char* name = variable_names[variable];
+  const char* value = std::getenv(name);
   if (!value) {
     return nullopt;
   }
   return StringView{value};
 }
 
+void set(Variable variable, StringView value) {
```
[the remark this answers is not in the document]
xlamorlette-datadog: I think we can indeed have a race condition, on Windows, between the getenv() and putenv() calls.

Actually, I'm a bit confused: What is the purpose of the 'root session id'? For example, for Nginx, is it to identify an Nginx instance?
I need time to study the [RFC](https://docs.google.com/document/d/1ECKj9_NnwaKYtFqm3p3Rlpicx5d-OQcdj9kI2jvRqVU).

Also, I think we should check the return values of `putenv()` and `setenv()`.

include/datadog/environment.h
khanayan123: Done

include/datadog/environment.h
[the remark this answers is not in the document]
khanayan123: Done

include/datadog/tracer_signature.h
khanayan123: Added a 3-param constructor that defaults root_session_id to runtime_id.string(), and simplified the init() call in telemetry.cpp to use it. Fuzz code also works without changes

test/telemetry/test_telemetry.cpp
```diff
@@ -353,6 +355,65 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
 }
 
+TELEMETRY_IMPLEMENTATION_TEST("session ID headers") {
+  auto logger = std::make_shared<MockLogger>();
+  auto client = std::make_shared<MockHTTPClient>();
+  auto scheduler = std::make_shared<FakeEventScheduler>();
+  auto url = HTTPClient::URL::parse("http://localhost:8000");
+
+  SECTION("root process: DD-Session-ID present, DD-Root-Session-ID absent") {
+    auto rid = RuntimeID::generate();
```
xlamorlette-datadog: _nit_: I suggest renaming `rid` in `session_rid` (or even `session_runtime_id`), to distinguish it clearly from the root one (same below).

include/datadog/tracer_signature.h
xlamorlette-datadog: _nit_: use the same names for the parameters than for the members (`id` → `runtime_id`, `root_session` → `root_session_id`, `service` → `default_service`, `environment` → `default_environment`).

test/test_datadog_agent.cpp
```diff
@@ -199,7 +199,8 @@ DATADOG_AGENT_TEST("Remote Configuration") {
   auto finalized = finalize_config(config);
   REQUIRE(finalized);
 
-  const TracerSignature signature(RuntimeID::generate(), "testsvc", "test");
+  auto rid = RuntimeID::generate();
+  const TracerSignature signature(rid, rid.string(), "testsvc", "test");
```
xlamorlette-datadog: _nit_: With the new three parameters constructor, you can go back to the previous version.

test/telemetry/test_telemetry.cpp
```diff
@@ -366,6 +368,65 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
+    const TracerSignature sig{rid, rid.string(), "testsvc", "test"};
```
xlamorlette-datadog: _nit_: You can use the three parameters constructor. Same below.

test/remote_config/test_remote_config.cpp
```diff
@@ -47,8 +47,10 @@ auto logger = std::make_shared<NullLogger>();
 
 REMOTE_CONFIG_TEST("initial state payload") {
   // Verify the initial payload structure for a remote configuration instance.
+  auto runtime_id = RuntimeID::generate();
   const TracerSignature tracer_signature{
-      /* runtime_id = */ RuntimeID::generate(),
+      /* runtime_id = */ runtime_id,
```
xlamorlette-datadog: _nit_: With the new three parameters constructor, you can go back to the previous version. Same below.

xlamorlette-datadog:
After reading the [RFC - Stable Service Instance Identifier Implementation Proposal](https://docs.google.com/document/d/1ECKj9_NnwaKYtFqm3p3Rlpicx5d-OQcdj9kI2jvRqVU), and after discussing with @zacharycmontoya, I agree with his proposal: it should simpler and more efficient to use a singleton rather than an environment variable.

The classic singleton implementation (with a `static` variable in the `static` method to get the instance) is thread safe (it is initialized once and only once by the first thread accessing it, without race issue). Also, since it should be immutable, it is safe to have it copied in the case of a fork.

Thus, I think that implementing the root session id with a singleton rather than with an environment variable should work in all cases.

khanayan123:
@xlamorlette-datadog Sounds good to clarify, the [parent RFC](https://docs.google.com/document/d/1radE-Ohh24H1ewtC5VpI-SKDfeLzA4nMeAQ-6JPxTM8/edit?usp=sharing) focuses on requiring the SDKs to report these headers. The propagation mechanism is an implementation detail left to each SDK's implementation. Since different SDKs have different concerns. Env var propagation seemed like the simplest cross-SDK approach for the cross-sdk implementation proposal but I agree when doable, a singleton is cleaner. I'll update the C++ implementation to use that instead.

src/datadog/telemetry/telemetry_impl.cpp
khanayan123: I agree app_started() could be simplified via send_payload(), but I'd prefer to keep that as a separate follow-up since it changes the on_response behavior (captures this for metric counters vs just logger) which breaks the CI's system tests and likely requires a larger refactor. Keeping this PR scoped to the session ID headers for now

src/datadog/root_session_id.h
```diff
@@ -0,0 +1,17 @@
...
+#include <string>
+
+namespace datadog {
+namespace tracing {
+namespace root_session_id {
+
+// Thread-safe singleton for the root session ID. The first call to `set`
+// captures the value; subsequent calls are no-ops. Immutable after init,
+// so fork-safe.
+void set(const std::string& id);
```
zacharycmontoya: I don't think we even need a `set` call. The `get` should populate the singleton so callers only have to call `get` and it will be returned in a thread-safe manner

zacharycmontoya: Oh I suppose that you have a `set` call because it might be set to the first `runtime-id` value. Is that right?

src/datadog/tracer_config.cpp
```diff
@@ -261,6 +262,11 @@ Expected<TracerConfig> load_tracer_env_config(Logger &logger) {
     return std::move(error);
   }
 
+  const auto &root_sid = root_session_id::get();
```
zacharycmontoya: Let's remove the get call from the configs now that we've determined it won't be an environment variable. This will allow us to have one clear `set` call in the Tracer and one clear `get` call when telemetry is reported

include/datadog/tracer_config.h
```diff
@@ -189,6 +189,10 @@ struct TracerConfig {
   // This option is ignored if `resource_renaming_enabled` is not `true`.
   Optional<bool> resource_renaming_always_simplified_endpoint;
 
+  // Populated internally from the root_session_id singleton. Tracks the
```
zacharycmontoya: In my other [comment](https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3042072101) I mentioned that we no longer need this in our config objects, so we can remove these changes as well

src/datadog/tracer.cpp
```diff
@@ -48,8 +49,9 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
     : logger_(config.logger),
       runtime_id_(config.runtime_id ? *config.runtime_id
                                     : RuntimeID::generate()),
-      signature_{runtime_id_, config.defaults.service,
-                 config.defaults.environment},
+      signature_{runtime_id_,
+                 config.root_session_id.value_or(runtime_id_.string()),
```
zacharycmontoya: This is going to sound really weird, but I have a suggestion for updating this line of code. I think Method 1 Meyers Singleton from this [Google AI response](https://share.google/aimode/g2WWAmV5HzDDZbndC) would be the best way to initialize the root session ID singleton. If we change the singleton implementation to align with that, then what we will be doing is:
1. We are guaranteed to have a  valid `runtime_id_` from the previous line
2. Right after we've set the `runtime_id_` field, we _try_ to initialize the Singleton root session ID with the `runtime_id_` value, but the returned value will always be the singleton from the first `runtime_id_` that was set
3. Once we get the singleton instance back, then we get the string value so we can assign it to the `root_session_id` field

```suggestion
                 Singleton::getInstance(runtime_id_.string()).getValue(),
```

src/datadog/root_session_id.h
zacharycmontoya: Resolving since I have a separate [suggestion](https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3042117849) for how we should implement the singleton

include/datadog/tracer_signature.h
```diff
@@ -33,17 +33,25 @@ namespace tracing {
...
   StringView library_language_version;
 
   TracerSignature() = delete;
-  TracerSignature(RuntimeID id, std::string service, std::string environment)
-      : runtime_id(id),
-        default_service(std::move(service)),
-        default_environment(std::move(environment)),
+  TracerSignature(RuntimeID runtime_id, std::string default_service,
+                  std::string default_environment)
+      : TracerSignature(runtime_id, runtime_id.string(),
```
zacharycmontoya, line 47: The logic for setting the root session ID should match what we do here, where we always _try_ to initialize the singleton with the given runtime_id: https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3042117849

test/telemetry/test_telemetry.cpp
```diff
@@ -366,6 +366,67 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
   }
 }
 
+TELEMETRY_IMPLEMENTATION_TEST("session ID headers") {
+  auto logger = std::make_shared<MockLogger>();
+  auto client = std::make_shared<MockHTTPClient>();
+  auto scheduler = std::make_shared<FakeEventScheduler>();
+  auto url = HTTPClient::URL::parse("http://localhost:8000");
```
zacharycmontoya, line 373: Depending on changes I've suggested [here](https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3042117849) you may need to acquire the `root_session_id` here and then use that throughout these test cases to exercise the cases where `session_id` and `root_session_id` do/do not match

include/datadog/tracer_signature.h
khanayan123: Just to confirm are you suggesting that the 3-param TracerSignature constructor itself should call root_session_id::get_or_init(runtime_id.string()) instead of defaulting to runtime_id.string()? Currently we do that in tracer.cpp at construction time, but the constructor itself stays decoupled from the singleton so tests/fuzz code can construct TracerSignature directly without going through the singleton

include/datadog/tracer_signature.h
```diff
@@ -33,17 +33,25 @@ namespace tracing {
...
   std::string library_version;
   StringView library_language;
   StringView library_language_version;
 
   TracerSignature() = delete;
-  TracerSignature(RuntimeID id, std::string service, std::string environment)
-      : runtime_id(id),
-        default_service(std::move(service)),
-        default_environment(std::move(environment)),
+  TracerSignature(RuntimeID runtime_id, std::string default_service,
```
xlamorlette-datadog, line 45: _nit, pedantic_: add a blank lines between the different constructors to improve readability.

include/datadog/tracer_signature.h
xlamorlette-datadog: Indeed, I don't think `TracerSignature`, which is a Data Transfer Object, should depend on `root_session_id`.

src/datadog/root_session_id.cpp
```diff
@@ -0,0 +1,33 @@
+#include "root_session_id.h"
+
+#include <atomic>
+
+namespace datadog {
```
xlamorlette-datadog: _nit_: rather use C++17 compound namespaces:
```
namespace datadog::tracing::root_session_id {
```

src/datadog/root_session_id.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+const std::string& get_or_init(const std::string& runtime_id) {
+  bool expected = false;
+  if (initialized().compare_exchange_strong(expected, true)) {
+    instance() = runtime_id;
+  }
+  return instance();
+}
+
+}  // namespace root_session_id
```
xlamorlette-datadog: _nit, pedantic, personal preference_: I think such comments are noise

src/datadog/root_session_id.cpp
```diff
@@ -0,0 +1,33 @@
...
+std::atomic<bool>& initialized() {
+  static std::atomic<bool> flag{false};
+  return flag;
+}
+
+}  // namespace
+
+const std::string& get_or_init(const std::string& runtime_id) {
+  bool expected = false;
+  if (initialized().compare_exchange_strong(expected, true)) {
```
xlamorlette-datadog: What do you need this, rather than implementing the singleton directly:
```
const std::string& get_or_init(const std::string& runtime_id) {
  static const std::string id = runtime_id;
  return id;
}
```

By using `inline` in the header file, you can avoid the whole `.cpp` file:
```
inline const std::string& get_or_init(const std::string& runtime_id) {
  static const std::string id = runtime_id;
  return id;
}
```

src/datadog/root_session_id.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <string>
+
+namespace datadog {
```
xlamorlette-datadog: _nit_: rather use C++17 compound namespaces:
```
namespace datadog::tracing::root_session_id {
```

src/datadog/root_session_id.h
```diff
@@ -0,0 +1,13 @@
...
+namespace tracing {
+namespace root_session_id {
+
+const std::string& get_or_init(const std::string& runtime_id);
+
+}  // namespace root_session_id
```
xlamorlette-datadog: _nit, pedantic, personal preference_: I think such comments are noise

src/datadog/telemetry/telemetry_impl.cpp
```diff
@@ -60,6 +60,15 @@ HTTPClient::URL make_telemetry_endpoint(HTTPClient::URL url) {
   return url;
 }
 
+void set_session_headers(DictWriter& headers,
+                         const tracing::TracerSignature& sig) {
```
xlamorlette-datadog: _nit, clean code_: use explicit names rather than uncommon abbreviation: `signature`
same line 322 and 378

src/datadog/tracer.cpp
```diff
@@ -48,8 +49,9 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
     : logger_(config.logger),
       runtime_id_(config.runtime_id ? *config.runtime_id
                                     : RuntimeID::generate()),
-      signature_{runtime_id_, config.defaults.service,
-                 config.defaults.environment},
+      signature_{runtime_id_,
```
xlamorlette-datadog, line 52: _nit_: I agree one should generalize the usage of braces initialization calls, but here I think we should favor consistency with existing code

test/telemetry/test_telemetry.cpp
```diff
@@ -366,6 +367,67 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
+  auto init_rid = RuntimeID::generate();
+  const auto& root_id = root_session_id::get_or_init(init_rid.string());
+
+  SECTION("root process: DD-Session-ID present, DD-Root-Session-ID absent") {
+    auto session_rid = RuntimeID::generate();
+    const TracerSignature sig{session_rid, session_rid.string(), "testsvc",
+                              "test"};
+
+    Telemetry telemetry{*finalize_config(), sig, logger, client,
+                        scheduler,          *url};
```
xlamorlette-datadog: _nit_: extra spaces; same below

test/telemetry/test_telemetry.cpp
```diff
@@ -366,6 +367,67 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
+  auto client = std::make_shared<MockHTTPClient>();
+  auto scheduler = std::make_shared<FakeEventScheduler>();
+  auto url = HTTPClient::URL::parse("http://localhost:8000");
+
+  auto init_rid = RuntimeID::generate();
+  const auto& root_id = root_session_id::get_or_init(init_rid.string());
+
+  SECTION("root process: DD-Session-ID present, DD-Root-Session-ID absent") {
+    auto session_rid = RuntimeID::generate();
+    const TracerSignature sig{session_rid, session_rid.string(), "testsvc",
```
xlamorlette-datadog: Use the 3 parameters constructor.

test/telemetry/test_telemetry.cpp
```diff
@@ -366,6 +367,67 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
 }
 
+TELEMETRY_IMPLEMENTATION_TEST("session ID headers") {
+  auto logger = std::make_shared<MockLogger>();
+  auto client = std::make_shared<MockHTTPClient>();
+  auto scheduler = std::make_shared<FakeEventScheduler>();
+  auto url = HTTPClient::URL::parse("http://localhost:8000");
+
+  auto init_rid = RuntimeID::generate();
+  const auto& root_id = root_session_id::get_or_init(init_rid.string());
```
xlamorlette-datadog: I think we should avoid calling the singleton in the unit tests.

xlamorlette-datadog (commented):
Please review the singleton implementation.

include/datadog/tracer_signature.h
zacharycmontoya: Hmm I see. Nevermind my suggestion, let's keep this behavior where we do not update the singleton inside this TracerSignature constructor

zacharycmontoya (approved):
LGTM. Thanks for incorporating all our feedback, this is a much more simplified implementation now! Please also get approval @xlamorlette-datadog before merging.

test/telemetry/test_telemetry.cpp
xlamorlette-datadog: up

xlamorlette-datadog (approved):

include/datadog/tracer_config.h
```diff
@@ -189,6 +189,11 @@ struct TracerConfig {
   // This option is ignored if `resource_renaming_enabled` is not `true`.
   Optional<bool> resource_renaming_always_simplified_endpoint;
 
+  // Root session ID for stable telemetry correlation across forked workers.
+  // Integrations (nginx, httpd, kong) should set this in the master process
+  // before workers fork so all Tracers share the same root.
+  Optional<std::string> root_session_id;
```
xlamorlette-datadog, line 195: _nit_: we should have the same order for the fields in `TracerConfig` and `FinalizedTracerConfig`.

src/datadog/tracer_config.cpp
```diff
@@ -407,6 +407,10 @@ Expected<FinalizedTracerConfig> finalize_config(const TracerConfig &user_config,
     final_config.runtime_id = user_config.runtime_id;
   }
 
+  if (user_config.root_session_id) {
```
xlamorlette-datadog: The test `if (user_config.root_session_id)` is useless since both fields are `Optional` and since `final_config` has just been instantiated with default values.
Same above for `  if (user_config.runtime_id)`.

src/datadog/tracer.cpp
xlamorlette-datadog: up

test/telemetry/test_telemetry.cpp
```diff
@@ -366,6 +366,67 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
+    auto session_it = client->request_headers.items.find("DD-Session-ID");
+    REQUIRE(session_it != client->request_headers.items.end());
+    CHECK(session_it->second == session_rid.string());
+
+    auto root_it = client->request_headers.items.find("DD-Root-Session-ID");
+    REQUIRE(root_it != client->request_headers.items.end());
+    CHECK(root_it->second == root_rid.string());
+  }
+
+  SECTION("heartbeat includes session headers") {
```
xlamorlette-datadog, line 408: I'm a bit puzzled: is is expected that the only difference with the `child process: DD-Root-Session-ID present when different` test is the following two calls:
```
    client->clear();
    scheduler->trigger_heartbeat();
```
But, after that, there is no difference in what is checked.

xlamorlette-datadog (commented):
lgtm

test/telemetry/test_telemetry.cpp
khanayan123: The test verifies headers persist on heartbeat requests (after client->clear() + trigger_heartbeat()), not just app-started. The assertions are intentionally identical same headers should appear on every telemetry event

khanayan123:
That's clang-format's alignment, CI's check-format step enforces it

https://github.com/DataDog/dd-trace-cpp/pull/295#discussion_r3064791971

xlamorlette-datadog:
@khanayan123 You forgot to solve the following comments (that you agreed):
- https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3064713138
- https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3046662390

khanayan123:
> @khanayan123 You forgot to solve the following comments (that you agreed):
> 
> * https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3064713138
> * https://github.com/DataDog/dd-trace-cpp/pull/295/changes#r3046662390

Hmm, must've got lost in one of the commits. Opening up a follow up

PR #304, opened 2026-04-02
Add backfilled option to config registry

BridgeAR:
The backfilled option showed some issue with the current state of
the supported-configurations.json. This is fixed by this PR by
adding setting backfilled to true as well as updating the generation
script to:
- lower case types
- use true and false as boolean
- use exact decimal values as defined in the macro

In addition, change DD_TRACE_RATE_LIMIT and
DD_TELEMETRY_EXTENDED_HEARTBEAT_INTERVAL decimal types to integers.

Some default values were changed from an empty string to `nullptr`, since
that seemed more suitable for these entries.

The script now also preserves the former implementation used.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -21,7 +21,7 @@ validate_supported_configurations_v2_local_file:
   extends: .validate_supported_configurations_v2_local_file
   variables:
     LOCAL_JSON_PATH: "supported-configurations.json"
-    BACKFILLED: false
+    BACKFILLED: true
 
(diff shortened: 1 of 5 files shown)


dmehala (approved):

dmehala:
@BridgeAR I am trying to enforce [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/). Can you make sure to follow the convention when you're going to merge?

zacharycmontoya (approved):

PR #309, opened 2026-04-17
feat(c-binding): accept explicit span timestamps

adeshkumar1:
## Summary
Adds `start_time_ns` to `dd_span_options_t` and `dd_span_set_end_time(span, end_time_ns)` so FFI callers (Kong) can pass wall-clock nanoseconds captured before the FFI crossing instead of letting the binding record "now." `start_time_ns = DD_TRACE_CURRENT_TIME` (-1) preserves the current behaviour; `dd_span_finish` is unchanged.

## Notes
- At span create, `start.tick` is derived from a current system/steady offset sample (the canonical C++17 pattern; C++20's `clock_cast` isn't available). In `dd_span_set_end_time`, `end.tick` is anchored to `start.tick + (end_wall − start_wall)` so mid-span clock adjustments (NTP, VM suspend) don't distort the caller's reported duration.
- End-time is clamped to start-time — serialization casts duration to `uint64_t`, so `end < start` would ship as a bogus huge value.
- Appending to `dd_span_options_t` is source-compatible (designated initializers zero-init the new field, which is then treated as the 1970-epoch); ABI grows by 8 bytes, so recompilation is required. Callers who want "use current time" must set `.start_time_ns = DD_TRACE_CURRENT_TIME` explicitly.

### DIFF
--- a/binding/c/include/datadog/c/tracer.h
+++ b/binding/c/include/datadog/c/tracer.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <stddef.h>
+#include <stdint.h>
 
 #if defined(_WIN32)
 #if defined(DD_TRACE_C_BUILDING)
@@ -51,6 +52,7 @@ typedef struct {
   const char* service_type;
   const char* environment;
   const char* version;
+  int64_t start_time_ns;  // UNIX-epoch nanoseconds; 0 = use current time
 } dd_span_options_t;
 
 // Error codes returned by the C binding.
@@ -178,6 +180,13 @@ DD_TRACE_C_API dd_span_t* dd_span_create_child(dd_span_t* span_handle,
 //
 // @param span_handle Span handle
 DD_TRACE_C_API void dd_span_finish(dd_span_t* span_handle);
+
+// Finish a span using an explicit end time. No-op if span_handle is NULL.
+//
+// @param span_handle Span handle
+// @param end_time_ns Wall-clock end time (UNIX-epoch ns); 0 = current time
+DD_TRACE_C_API void dd_span_finish_with_time(dd_span_t* span_handle,
+                                             int64_t end_time_ns);
 
 // Get the trace ID as a zero-padded hex string.
 //
--- a/binding/c/include/datadog/c/tracer.h
+++ b/binding/c/include/datadog/c/tracer.h
@@ -44,6 +44,9 @@ typedef enum {
   DD_OPT_INTEGRATION_VERSION = 5
 } dd_tracer_option;
 
+// Sentinel for start_time_ns meaning "use the current time."
+static const int64_t DD_TRACE_CURRENT_TIME = -1;
+
 // Options for creating a span. Unset fields default to NULL.
 typedef struct {
   const char* name;
@@ -52,7 +55,7 @@ typedef struct {
   const char* service_type;
   const char* environment;
   const char* version;
-  int64_t start_time_ns;  // UNIX-epoch nanoseconds; 0 = use current time
+  int64_t start_time_ns;  // Unix-epoch nanoseconds or DD_TRACE_CURRENT_TIME.
 } dd_span_options_t;
 
 // Error codes returned by the C binding.
@@ -181,12 +184,17 @@ DD_TRACE_C_API dd_span_t* dd_span_create_child(dd_span_t* span_handle,
 // @param span_handle Span handle
 DD_TRACE_C_API void dd_span_finish(dd_span_t* span_handle);
 
-// Finish a span using an explicit end time. No-op if span_handle is NULL.
+// Set the end time of a span from an explicit wall-clock timestamp. The
+// span itself is finished when dd_span_free destroys it. To use the
+// current time, call dd_span_finish instead. If end_time_ns is earlier
+// than the span's start time, the resulting duration is clamped to zero.
+// No-op if span_handle is NULL.
 //
 // @param span_handle Span handle
-// @param end_time_ns Wall-clock end time (UNIX-epoch ns); 0 = current time
-DD_TRACE_C_API void dd_span_finish_with_time(dd_span_t* span_handle,
-                                             int64_t end_time_ns);
+// @param end_time_ns Wall-clock end time (Unix-epoch nanoseconds).
+//                    DD_TRACE_CURRENT_TIME does not apply here.
+DD_TRACE_C_API void dd_span_set_end_time(dd_span_t* span_handle,
+                                         int64_t end_time_ns);
 
 // Get the trace ID as a zero-padded hex string.
 //
--- a/binding/c/src/tracer.cpp
+++ b/binding/c/src/tracer.cpp
@@ -44,6 +44,18 @@ class ContextWriter : public dd::DictWriter {
   }
 };
 
+// tick is derived from the current system/steady offset so that duration
+// (end.tick - start.tick) stays correct across FFI-supplied timestamps.
+dd::TimePoint wall_ns_to_timepoint(int64_t wall_ns) {
+  const auto wall = std::chrono::system_clock::time_point(
+      std::chrono::round<std::chrono::system_clock::duration>(
+          std::chrono::nanoseconds(wall_ns)));
+  const auto now_wall = std::chrono::system_clock::now();
+  const auto now_tick = std::chrono::steady_clock::now();
+  return {wall,
+          now_tick - std::chrono::duration_cast<dd::Duration>(now_wall - wall)};
+}
+
 dd::SpanConfig make_span_config(dd_span_options_t options) {
   dd::SpanConfig span_config;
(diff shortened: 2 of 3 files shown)


adeshkumar1:
@codex review

adeshkumar1:
@codex review

adeshkumar1:
@codex review

adeshkumar1:
@codex review

binding/c/src/tracer.cpp
```diff
@@ -249,12 +264,33 @@ dd_span_t *dd_span_create_child(dd_span_t *span_handle,
...
+  std::chrono::steady_clock::time_point end_tick;
+  if (end_time_ns == 0) {
+    end_tick = std::chrono::steady_clock::now();
+  } else {
+    const auto end_wall = std::chrono::system_clock::time_point(
+        std::chrono::round<std::chrono::system_clock::duration>(
+            std::chrono::nanoseconds(end_time_ns)));
+    end_tick = start.tick + (end_wall > start.wall
+                                 ? std::chrono::duration_cast<dd::Duration>(
+                                       end_wall - start.wall)
```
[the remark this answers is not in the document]
adeshkumar1: Intentionally keeping the current `start.tick + (end_wall - start.wall)` anchoring. The previous review flagged the opposite behaviour (re-sampling the system/steady offset at finish) as a P1 because it would corrupt duration when the system clock shifted mid-span. Swapping to re-sample would re-introduce that bug.

For this API's use case — FFI callers capturing wall-clock timestamps at lifecycle boundaries (e.g. Kong reading nginx's `ngx_time_update()` → `gettimeofday`) — the caller's wall-clock delta IS what they want reported as duration. A clock step during the span is already baked into the two timestamps they supplied; showing 70s for a span that the caller observes as 70s of wall time (including a VM suspend) is the honest answer, not a bug.

If a future caller wants steady-clock-anchored duration, they'd pass `start_time_ns=0` and `end_time_ns=0` to use the default clock — which is the existing, always-available code path.

binding/c/include/datadog/c/tracer.h
```diff
@@ -179,6 +181,13 @@ DD_TRACE_C_API dd_span_t* dd_span_create_child(dd_span_t* span_handle,
 // @param span_handle Span handle
 DD_TRACE_C_API void dd_span_finish(dd_span_t* span_handle);
 
+// Finish a span using an explicit end time. No-op if span_handle is NULL.
+//
+// @param span_handle Span handle
+// @param end_time_ns Wall-clock end time (UNIX-epoch ns); 0 = current time
+DD_TRACE_C_API void dd_span_finish_with_time(dd_span_t* span_handle,
```
dmehala: I suggest:
```suggestion
DD_TRACE_C_API void dd_span_set_endtime(dd_span_t* span_handle,
```

to follow the current span API: https://github.com/DataDog/dd-trace-cpp/blob/1f76c5cd2c8c9f8a472643b24f559b44be1fac12/include/datadog/span.h#L165

binding/c/include/datadog/c/tracer.h
```diff
@@ -51,6 +52,7 @@ typedef struct {
   const char* service_type;
   const char* environment;
   const char* version;
+  int64_t start_time_ns;  // UNIX-epoch nanoseconds; 0 = use current time
```
dmehala: I'd suggest to use `-1` instead and expose a `const` variable:
```suggestion
  const int64_t DD_TRACE_CURRENT_TIME = -1;

  int64_t` start_time_ns;  // UNIX-epoch nanoseconds; -1 = use current time
```

binding/c/src/tracer.cpp
```diff
@@ -250,11 +265,33 @@ dd_span_t *dd_span_create_child(dd_span_t *span_handle,
...
-      ->set_end_time(std::chrono::steady_clock::now());
+  auto *span = reinterpret_cast<dd::Span *>(span_handle);
+  const auto start = span->start_time();
+  // Explicit path anchors to start.wall so NTP adjustments during the
+  // span's lifetime don't skew duration; implicit path uses steady now.
+  // Both clamp to start.tick so a negative duration never ships
+  // (serialization casts duration to uint64_t).
+  std::chrono::steady_clock::time_point end_tick;
+  if (end_time_ns == 0) {
+    end_tick = std::chrono::steady_clock::now();
```
dmehala: if `end_time_ns` the c++ layer will handle this case.

binding/c/include/datadog/c/tracer.h
```diff
@@ -43,6 +44,9 @@ typedef enum {
   DD_OPT_INTEGRATION_VERSION = 5
 } dd_tracer_option;
 
+// Sentinel for start_time_ns meaning "use the current time."
+static const int64_t DD_TRACE_CURRENT_TIME = -1;
```
xlamorlette-datadog, line 48: This is incoherent with the PR description:
> 0 on either field preserves the current behaviour, so existing callers and other language bindings are unaffected.

binding/c/include/datadog/c/tracer.h
```diff
@@ -51,6 +55,7 @@ typedef struct {
   const char* service_type;
   const char* environment;
   const char* version;
+  int64_t start_time_ns;  // UNIX-epoch nanoseconds or DD_TRACE_CURRENT_TIME.
```
xlamorlette-datadog: _nit, opt, ultra pedantic_: casing: `Unix`
same line 192

binding/c/src/tracer.cpp
```diff
@@ -44,6 +44,18 @@ class ContextWriter : public dd::DictWriter {
   }
 };
 
+// tick is derived from the current system/steady offset so that duration
+// (end.tick - start.tick) stays correct across FFI-supplied timestamps.
```
xlamorlette-datadog: I find confusing to talk about end.tick and start.tick here: I think this relates to how this function is used in the caller code, it does not relate to the function in its own.
I suggest the following comment (to be amended if necessary):
```
Create a TimePoint from a wall clock timestamp in nanoseconds.
The tick is derived from the wall clock using the current offset between the system and the steady clocks.
This ensures computing a duration from ticks remains accurate when the start time is supplied via FFI.
```

binding/c/src/tracer.cpp
```diff
@@ -257,6 +272,25 @@ void dd_span_finish(dd_span_t *span_handle) {
...
+void dd_span_set_end_time(dd_span_t *span_handle, int64_t end_time_ns) {
+  if (span_handle == nullptr) {
+    return;
+  }
+  auto *span = reinterpret_cast<dd::Span *>(span_handle);
+  // Anchor end.tick to start.tick + wall delta so NTP adjustments during the
+  // span's lifetime don't skew duration. Clamp to start.tick on negative
+  // deltas — serialization casts duration to uint64_t.
+  const auto start = span->start_time();
+  const auto end_wall = std::chrono::system_clock::time_point(
```
xlamorlette-datadog: _nit_: We could factorize code with what is in `wall_ns_to_timepoint()`:
```
std::chrono::system_clock::time_point wall_ns_to_system_time(int64_t wall_ns) {
  return std::chrono::system_clock::time_point(
      std::chrono::round<std::chrono::system_clock::duration>(
          std::chrono::nanoseconds(wall_ns)));
}
```

binding/c/include/datadog/c/tracer.h
```diff
@@ -179,6 +184,15 @@ DD_TRACE_C_API dd_span_t* dd_span_create_child(dd_span_t* span_handle,
 // @param span_handle Span handle
 DD_TRACE_C_API void dd_span_finish(dd_span_t* span_handle);
 
+// Set the end time of a span from an explicit wall-clock timestamp. The
+// span itself is finished when dd_span_free destroys it. To use the
+// current time, call dd_span_finish instead. No-op if span_handle is NULL.
+//
+// @param span_handle Span handle
+// @param end_time_ns Wall-clock end time (UNIX-epoch ns)
```
xlamorlette-datadog: _nit_: maybe worth mentioning that `DD_TRACE_CURRENT_TIME` should not be used here

binding/c/include/datadog/c/tracer.h
```diff
@@ -179,6 +184,15 @@ DD_TRACE_C_API dd_span_t* dd_span_create_child(dd_span_t* span_handle,
 // @param span_handle Span handle
 DD_TRACE_C_API void dd_span_finish(dd_span_t* span_handle);
 
+// Set the end time of a span from an explicit wall-clock timestamp. The
+// span itself is finished when dd_span_free destroys it. To use the
+// current time, call dd_span_finish instead. No-op if span_handle is NULL.
+//
```
xlamorlette-datadog, line 192: We should add a comment about clamping duration to zero if the provided end time is before the start.

xlamorlette-datadog (approved):
Please review my comments, and I think we're good to go.

PR #311, opened 2026-04-23
benchmark: add micro-benchmarks for TraceID hex formatting and parsing

AneesPatel:
# Description
Adds `benchmark/trace_id_bench.cpp` to introduce targeted Google Benchmark coverage for the 128-bit and 64-bit TraceID formatting paths. Also updates `benchmark/CMakeLists.txt` to include the new benchmark executable.

**Baseline Metrics (Ubuntu 22.04 / GCC 11.4):**
| Benchmark | Time (ns/op) | Iterations | Notes |
|---|---|---|---|
| `BM_TraceID_HexPadded` | 204 ns | 3.6M | Full 128-bit ID to 32-char hex. Cost reflects two underlying `std::string` allocations and concatenation. |
| `BM_TraceID_ParseHex_128bit` | 108 ns | 6.5M | W3C `traceparent` extraction path. |
| `BM_TraceID_ParseHex_64bit` | 60 ns | 11.8M | Legacy 64-bit extraction path. |
| `BM_HexPadded_uint64` | 46 ns | 17.6M | Single `to_chars` + `std::string` alloc. |
| `BM_Hex_uint64` | 24 ns | 29.5M | No padding. |

# Motivation
`TraceID::hex_padded()` is called on every outgoing W3C `traceparent` header injection, and `TraceID::parse_hex()` on every incoming extraction. Because these functions sit in the hottest path of distributed tracing, their latency and allocation overhead multiply across millions of spans.

Currently, the underlying `hex_padded<uint64_t>()` and `hex<uint64_t>()` functions utilize a stack-buffer-to-`std::string` pattern. While the existing `BM_TraceTinyCCSource` benchmark provides excellent end-to-end throughput data, it cannot isolate micro-regressions in these specific serialization operations. 

This baseline provides the data needed to safely prevent future regressions and evaluate allocation-reduction strategies in the critical path.

# Jira ticket 
N/A

# Additional Notes
- Validated locally via Linux (WSL2 Ubuntu 22.04) with GCC 11.4.

## How to run

```sh
mkdir .build && cd .build
cmake .. -DDD_TRACE_BUILD_BENCHMARK=1 -DCMAKE_BUILD_TYPE=Release
make -j
./benchmark/dd_trace_cpp-benchmark --benchmark_filter='^BM_TraceID|^BM_Hex'

### DIFF
--- a/benchmark/trace_id_bench.cpp
+++ b/benchmark/trace_id_bench.cpp
@@ -1,0 +1,54 @@
+#include <benchmark/benchmark.h>
+#include <datadog/trace_id.h>
+
+#include "datadog/hex.h"
+
+namespace {
+namespace dd = datadog::tracing;
+
+void BM_TraceID_HexPadded(benchmark::State& state) {
+  const dd::TraceID id{0xDEADBEEFCAFEBABEULL, 0x0102030405060708ULL};
+  for (auto _ : state) {
+    auto result = id.hex_padded();
+    benchmark::DoNotOptimize(result);
+  }
+}
+BENCHMARK(BM_TraceID_HexPadded);
+
+void BM_TraceID_ParseHex_128bit(benchmark::State& state) {
+  const std::string input{"0102030405060708deadbeefcafebabe"};
+  for (auto _ : state) {
+    auto result = dd::TraceID::parse_hex(input);
+    benchmark::DoNotOptimize(result);
+  }
+}
+BENCHMARK(BM_TraceID_ParseHex_128bit);
+
+void BM_TraceID_ParseHex_64bit(benchmark::State& state) {
+  const std::string input{"deadbeefcafebabe"};
+  for (auto _ : state) {
+    auto result = dd::TraceID::parse_hex(input);
+    benchmark::DoNotOptimize(result);
+  }
+}
+BENCHMARK(BM_TraceID_ParseHex_64bit);
+
+void BM_HexPadded_uint64(benchmark::State& state) {
+  const std::uint64_t value = 0xDEADBEEFCAFEBABEULL;
+  for (auto _ : state) {
+    auto result = dd::hex_padded(value);
+    benchmark::DoNotOptimize(result);
+  }
+}
+BENCHMARK(BM_HexPadded_uint64);
+
+void BM_Hex_uint64(benchmark::State& state) {
+  const std::uint64_t value = 0xDEADBEEFCAFEBABEULL;
+  for (auto _ : state) {
+    auto result = dd::hex(value);
+    benchmark::DoNotOptimize(result);
+  }
+}
+BENCHMARK(BM_Hex_uint64);
+
+}  // namespace
--- a/benchmark/trace_id_bench.cpp
+++ b/benchmark/trace_id_bench.cpp
@@ -6,41 +6,34 @@
 namespace {
 namespace dd = datadog::tracing;
 
-void BM_TraceID_HexPadded(benchmark::State& state) {
-  const dd::TraceID id{0xDEADBEEFCAFEBABEULL, 0x0102030405060708ULL};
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: benchmark/trace_id_bench.cpp [hunk 1]
```diff
@@ -0,0 +1,54 @@
+#include <benchmark/benchmark.h>
+#include <datadog/trace_id.h>
+
+#include "datadog/hex.h"
+
+namespace {
+namespace dd = datadog::tracing;
+
+void BM_TraceID_HexPadded(benchmark::State& state) {
+  const dd::TraceID id{0xDEADBEEFCAFEBABEULL, 0x0102030405060708ULL};
```
xlamorlette-datadog: Please add another test with values testing the padding path, id est with not all nibbles being zero, for example `TraceID{1, 0}`.
Similarly, please add a test for `HexPadded_uint64` with a “worst case” value such as`0x1ull`.

benchmark/trace_id_bench.cpp
```diff
@@ -0,0 +1,54 @@
...
+void BM_TraceID_ParseHex_128bit(benchmark::State& state) {
+  const std::string input{"0102030405060708deadbeefcafebabe"};
+  for (auto _ : state) {
+    auto result = dd::TraceID::parse_hex(input);
+    benchmark::DoNotOptimize(result);
+  }
+}
+BENCHMARK(BM_TraceID_ParseHex_128bit);
+
+void BM_TraceID_ParseHex_64bit(benchmark::State& state) {
```
xlamorlette-datadog: _nit_: `BM_TraceID_ParseHex_64bit()` and `BM_TraceID_ParseHex_128bit()` can be factorized as follows:
```
void BM_TraceID_ParseHex(benchmark::State& state, std::string input) {
  for (auto _ : state) {
    auto result = dd::TraceID::parse_hex(input);
    benchmark::DoNotOptimize(result);
  }
}
BENCHMARK_CAPTURE(BM_TraceID_ParseHex, 64bit,
                  std::string{"deadbeefcafebabe"});
BENCHMARK_CAPTURE(BM_TraceID_ParseHex, 128bit,
                  std::string{"0102030405060708deadbeefcafebabe"});
```

Such factorization technique can also be used for the additional values to test suggested in the comment above.

xlamorlette-datadog (commented):
Please, what is you personal objective in submitting this PR?

AneesPatel:
Hi @xlamorlette-datadog, thanks for the review and the great catch on the padding paths!

To answer your question, I'm currently an undergraduate studying computer science with an interest in low-latency systems and performance-oriented C++. My goal with this PR is to add measurable coverage for these serialization paths so that future allocation-reduction work (e.g., using std::to_chars with stack-allocated buffers to avoid the std::string heap allocation) can be evaluated with concrete latency data instead of guesswork.

I completely agree with testing the edge cases, and I've just pushed a commit that factorizes the tests using BENCHMARK_CAPTURE exactly as you suggested.

One interesting finding from my local runs is that BM_TraceID_HexPadded/WithPadding ({1, 0}) actually ran faster than the NoPadding case (~88 ns vs ~131 ns on my machine). It looks like when high == 0 the implementation skips formatting the high half entirely, which is why the WithPadding case ends up faster despite the name.

Let me know if the new structure looks good, thanks again for the review.

xlamorlette-datadog:
Some CI jobs are failing because this comes from a fork. I'm going to mirror this PR in our main repository.

xlamorlette-datadog:
All CI jobs are green on the [mirrored (draft) PR](https://github.com/DataDog/dd-trace-cpp/pull/311).

xlamorlette-datadog (approved):

xlamorlette-datadog:
@AneesPatel Would you please be able to sign your commits?
(This is blocking the PR merge.)

xlamorlette-datadog:
@AneesPatel Thanks for your contribution!

PR #313, opened 2026-05-04
fix: observe baggage limits on extraction

zacharycmontoya:
# Description
Implements the baggage limits on context extraction. Previously the limits only applied on context injection. This PR also applies context extraction limits on the `dd` tracestate information.

# Motivation
We want to implement a reasonable upper-bound on the size of incoming propagation headers.

# Additional Notes
N/A

### DIFF
--- a/examples/http-server/common/httplib.h
+++ b/examples/http-server/common/httplib.h
@@ -3110,7 +3110,7 @@ inline unsigned int str2tag(const std::string &s) {
 
 namespace udl {
 
-inline constexpr unsigned int operator"" _t(const char *s, size_t l) {
+inline constexpr unsigned int operator""_t(const char *s, size_t l) {
   return str2tag_core(s, l, 0);
 }
 
--- a/include/datadog/baggage.h
+++ b/include/datadog/baggage.h
@@ -57,8 +57,10 @@ class Baggage {
   /// instance if no errors are encounters .
   ///
   /// @param `reader` The input `DictReader` from which to extract the data.
+  /// @param `opts` Extraction options governing maximum items and bytes.
   /// @return A `Baggage` instance or an `Error`.
-  static Expected<Baggage, Error> extract(const DictReader& reader);
+  static Expected<Baggage, Error> extract(
+      const DictReader& reader, const Options& opts = default_options);
 
   /// Initializes an empty Baggage with the default maximum capacity.
   Baggage() = default;
--- a/src/datadog/baggage.cpp
+++ b/src/datadog/baggage.cpp
@@ -99,8 +99,9 @@ parse_baggage(StringView input, const Baggage::Options& opts) {
         if (c == '=') {
         consume_key:
           size_t count = tmp_end - beg;
-          if (count < 1)
+          if (count < 1) {
             return Baggage::Error{Baggage::Error::MALFORMED_BAGGAGE_HEADER, i};
+          }
 
           key = StringView{input.data() + beg, count};
           internal_state = state::leading_spaces_value;
@@ -144,15 +145,15 @@ parse_baggage(StringView input, const Baggage::Options& opts) {
         if (c == ',') {
         consume_value:
           size_t count = tmp_end - beg;
-          if (count < 1)
+          if (count < 1) {
             return Baggage::Error{Baggage::Error::MALFORMED_BAGGAGE_HEADER,
                                   tmp_end};
+          }
 
           value = StringView{input.data() + beg, count};
           if (result.size() >= opts.max_items) {
             return Baggage::Error{Baggage::Error::MAXIMUM_CAPACITY_REACHED};
           }
-
           result.emplace(std::string(key), std::string(value));
           beg = i;
           tmp_end = i;
@@ -179,7 +180,6 @@ parse_baggage(StringView input, const Baggage::Options& opts) {
     if (result.size() >= opts.max_items) {
       return Baggage::Error{Baggage::Error::MAXIMUM_CAPACITY_REACHED};
     }
-
     result.emplace(std::string(key), std::string(value));
   } else {
     return Baggage::Error{Baggage::Error::MALFORMED_BAGGAGE_HEADER, end};
--- a/src/datadog/baggage.cpp
+++ b/src/datadog/baggage.cpp
@@ -41,9 +41,15 @@ constexpr bool is_allowed_value_char(char c) {
 }
 
 Expected<std::unordered_map<std::string, std::string>, Baggage::Error>
-parse_baggage(StringView input) {
+parse_baggage(StringView input, const Baggage::Options& opts) {
   std::unordered_map<std::string, std::string> result;
   if (input.empty()) return result;
+
+  // This API throws an error when limits are exceeded, so we can simply check
+  // the entire header length against the maximum bytes limit.
+  if (input.size() > opts.max_bytes) {
+    return Baggage::Error{Baggage::Error::MAXIMUM_BYTES_REACHED};
+  }
 
   enum class state : char {
     leading_spaces_key,
@@ -143,6 +149,10 @@ parse_baggage(StringView input) {
                                   tmp_end};
 
           value = StringView{input.data() + beg, count};
+          if (result.size() >= opts.max_items) {
+            return Baggage::Error{Baggage::Error::MAXIMUM_CAPACITY_REACHED};
+          }
+
           result.emplace(std::string(key), std::string(value));
           beg = i;
           tmp_end = i;
@@ -158,10 +168,18 @@ parse_baggage(StringView input) {
 
   if (internal_state == state::value) {
     value = StringView{input.data() + beg, end - beg};
+    if (result.size() >= opts.max_items) {
+      return Baggage::Error{Baggage::Error::MAXIMUM_CAPACITY_REACHED};
+    }
+
     result.emplace(std::string(key), std::string(value));
   } else if (internal_state == state::trailing_spaces_value ||
              internal_state == state::properties) {
     value = StringView{input.data() + beg, tmp_end - beg};
+    if (result.size() >= opts.max_items) {
+      return Baggage::Error{Baggage::Error::MAXIMUM_CAPACITY_REACHED};
+    }
+
     result.emplace(std::string(key), std::string(value));
   } else {
     return Baggage::Error{Baggage::Error::MALFORMED_BAGGAGE_HEADER, end};
@@ -264,14 +282,15 @@ Expected<void> Baggage::inject(DictWriter& writer, const Options& opts) const {
   return res;
 }
 
-Expected<Baggage, Baggage::Error> Baggage::extract(const DictReader& headers) {
+Expected<Baggage, Baggage::Error> Baggage::extract(const DictReader& headers,
+                                                   const Options& opts) {
   auto found = headers.lookup("baggage");
   if (!found) {
     return Baggage::Error{Error::MISSING_HEADER};
   }
 
   // TODO(@dmehala): Avoid allocation
-  auto bv = parse_baggage(*found);
+  auto bv = parse_baggage(*found, opts);
   if (auto error = bv.if_error()) {
     return *error;
   }
--- a/src/datadog/extraction_util.cpp
+++ b/src/datadog/extraction_util.cpp
@@ -152,7 +152,20 @@ Expected<ExtractedData> extract_datadog(
 
   auto trace_tags = headers.lookup("x-datadog-tags");
   if (trace_tags) {
-    handle_trace_tags(*trace_tags, result, span_tags, logger);
+    // If the X-Datadog-Tags value exceeds 512 bytes, drop it and record a propagation error tag.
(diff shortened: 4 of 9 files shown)


examples/http-server/common/httplib.h
```diff
@@ -3110,7 +3110,7 @@ inline unsigned int str2tag(const std::string &s) {
 
 namespace udl {
 
-inline constexpr unsigned int operator"" _t(const char *s, size_t l) {
+inline constexpr unsigned int operator""_t(const char *s, size_t l) {
```
zacharycmontoya, line 3113: FYI: This was a style issue flagged by our formatter when building and running local tests

src/datadog/baggage.cpp
```diff
@@ -143,6 +149,10 @@ parse_baggage(StringView input) {
                                   tmp_end};
 
           value = StringView{input.data() + beg, count};
+          if (result.size() >= opts.max_items) {
```
zacharycmontoya, line 154: FYI: In this `state::trailing_spaces_value` part of the state machine, we have concluded that we have a baggage item to add to our collection, so throw an error if we would exceed the configured limit

src/datadog/baggage.cpp
```diff
@@ -158,10 +168,18 @@ parse_baggage(StringView input) {
 
   if (internal_state == state::value) {
```
zacharycmontoya, line 170: FYI: Similarly, if we finish parsing the input header and we end in this `state::value` part of the state machine, we have concluded that we have a baggage item to add to our collection, so throw an error if we would exceed the configured limit

src/datadog/w3c_propagation.cpp
```diff
@@ -299,6 +300,12 @@ void extract_tracestate(ExtractedData& result, const DictReader& headers) {
     result.additional_w3c_tracestate = std::move(other_entries);
   }
 
+  // If the "dd" vendor entry's value exceeds 512 bytes, drop it and record a propagation error tag.
+  if (datadog_value.size() > 512) {
```
zacharycmontoya, line 306: FYI: We only parse and manipulate our section of the W3C tracestate header, so always pass through the entries from other vendors but parse our portion or discard it if it exceeds the configured limits

awforsythe (approved):
(Apologies for my verbosity, but I like to talk through the thought process when I'm reviewing urgent changes in codebases I don't know well)

### baggage.h / .cpp

We're introducing two separate limits, `max_bytes` (total payload size) and `max_items` (total number of key/value pairs parsed). These limits are now enforced in `parse_baggage`.

`max_bytes` gets checked up-front with a simple length check on the input string: if the header value exceeds the maximum size, we reject it outright and perform no parsing. Simple enough.

`max_items` gets checked any time we're about to push a newly-parsed value into the `result` map. There are three existing places where the function calls `result.emplace` (and it doens't appear to mutate `result` in any other way, so this list oughta be exhaustive):

1. In the `consume_value` branch of the state machine logic, after we've reached the end of a value in the string
2. In the post-state-machine code that handles the last value, if we've concluded parsing in the `value` state, meaning we have a final value to add to the map
3. In the same code, just in the `trailing_spaces_value || properties` state, which has slightly different arithmetic for grabbing the portion of the string that represents the value, but is otherwise duplicated (duplication which is extremely minor and would be silly to refactor in an urgent bug-fix change)

...so in each of those three places, we now have preemptive checks for `result.size() >= max_items`. This will correctly allow the case where we have `max_items` values, and abort with an error in any conceivable case where we could otherwise accumulate more than `max_items`.

Validation logic looks correct to me.

`parse_baggage` is internal to `baggage.cpp`, so the extra parameter is fine. `Baggage::extract` is public, but the newly-added parameter is declared with a default so it won't break usage. The Codex review note re: this nevertheless breaking linkage seems worth considering, but I don't know what level of binary compatibility we're concerned with here, so I'll leave that to you.

### extraction_util.cpp

This looks like an entirely separate limit, and presumably `X-Datadog-Tags` is something that the library sets internally, hence the hardcoded, non-configurable limit. I'm guessing that exceeding 512 bytes would be unusual, and would indicate either a bug in the library or (more likely) the user configuring excessively long values.

Anyway, yup, logic is straightforward and looks good.

### tracer.cpp

We're just passing `baggage_opts_` to `Baggage::extract`. `baggage_opts_` is preexisting. OK.

### w3c_propagation.cpp

`extract_tracestate` (internal) gets a new parameter so that we can mutate `span_tags` if validation fails.

We have another separate, hardcoded 512-byte check, this time for `datadog_value`. Again, presumably internal-only and so a reasonable upper-bound-of-sanity to hardcode. Again, very simple logic to early-out if we exceed the max length.

### test/

Tests look sane; test coverage on `Baggage::extract` looks good. It might be nice to ensure that there are enough test cases to cover both the `value` and  `trailing_spaces_value || properties` branches for accumulation of the final value, but that's a nitpick.

Other tests look good.

(Sidenote: I've been doing the `GENERATE(values<TestCase>)` pattern but in a stupid way that uses `DYNAMIC_SECTION`; I knew there had to be a better way to do that. Thanks for the good example!)

---

As long as all of my stated assumptions line up with reality, LGTM.

include/datadog/baggage.h
```diff
@@ -57,8 +57,10 @@ class Baggage {
   /// instance if no errors are encounters .
   ///
   /// @param `reader` The input `DictReader` from which to extract the data.
+  /// @param `opts` Extraction options governing maximum items and bytes.
   /// @return A `Baggage` instance or an `Error`.
-  static Expected<Baggage, Error> extract(const DictReader& reader);
+  static Expected<Baggage, Error> extract(
+      const DictReader& reader, const Options& opts = default_options);
```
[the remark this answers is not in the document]
zacharycmontoya: We generally expect applications and web servers to re-compile with the updated dd-trace-cpp dependency, so this should not be an issue. I'll keep this comment open to track this observation

zacharycmontoya:
Thanks Alex! Yes all of the assumptions and reasoning are accurate, thanks for providing a review. I'll re-run CI (seems to be flaky GitHub Actions runners / Docker build times) and then it should be good to merge

src/datadog/baggage.cpp
```diff
@@ -143,6 +149,10 @@ parse_baggage(StringView input) {
                                   tmp_end};
 
           value = StringView{input.data() + beg, count};
+          if (result.size() >= opts.max_items) {
+            return Baggage::Error{Baggage::Error::MAXIMUM_CAPACITY_REACHED};
+          }
+
```
xlamorlette-datadog: _nit, opt, pedantic_: extra blank line (same below, lines 174 and 182)

src/datadog/baggage.cpp
xlamorlette-datadog, line 149: _opt_: By the way, could you please add braces around this line? Same line 103.

src/datadog/extraction_util.cpp
```diff
@@ -152,7 +152,21 @@ Expected<ExtractedData> extract_datadog(
 
   auto trace_tags = headers.lookup("x-datadog-tags");
   if (trace_tags) {
-    handle_trace_tags(*trace_tags, result, span_tags, logger);
+    // If the X-Datadog-Tags value exceeds 512 bytes, drop it and record a
```
xlamorlette-datadog: _nit, pedantic_: `x-datadog-tags` rather than `X-Datadog-Tags`

src/datadog/extraction_util.cpp
```diff
@@ -152,7 +152,21 @@ Expected<ExtractedData> extract_datadog(
...
+    // propagation error tag.
+    if (trace_tags->size() > 512) {
+      std::string message;
```
xlamorlette-datadog: _nit_: a bit terser:
```
logger.log_error([&](auto& stream) {
  stream << "Received x-datadog-tags header value is too large. The "
            "maximum size is 512 bytes, but the received value is "
         << trace_tags->size() << " bytes.";
});
```
(This pattern is, for example, already used in https://github.com/DataDog/dd-trace-cpp/blob/main/src/datadog/datadog_agent.cpp#L315.)

src/datadog/w3c_propagation.cpp
```diff
@@ -299,6 +301,13 @@ void extract_tracestate(ExtractedData& result, const DictReader& headers) {
     result.additional_w3c_tracestate = std::move(other_entries);
   }
 
+  // If the "dd" vendor entry's value exceeds 512 bytes, drop it and record a
+  // propagation error tag.
+  if (datadog_value.size() > 512) {
```
xlamorlette-datadog, line 306: _opt, nit_: We could we put 512, which is also in extraction_utils.cpp line 158, in a shared named constant. For example, `constexpr int propagation_tags_max_length = 512;` in a new `src/datadog/propagation_limits.h` file. But that may be overkill.

test/test_trace_segment.cpp
```diff
@@ -175,8 +175,8 @@ TEST_CASE("TraceSegment finalization of spans") {
 
   SECTION("root span") {
     SECTION(
-        "'inject_max_size' propagation error if X-Datadog-Tags oversized on "
```
xlamorlette-datadog, line 178: _nit_: casing: `x-datadog-tags`

test/test_baggage.cpp
```diff
@@ -163,6 +163,81 @@ BAGGAGE_TEST("extract") {
...
+    Baggage::Options opts{test_case.max_bytes, test_case.max_items};
+    auto maybe_baggage = Baggage::extract(reader, opts);
+    if (maybe_baggage.has_value() && test_case.expected_baggage.has_value()) {
+      CHECK(*maybe_baggage == *test_case.expected_baggage);
+    } else if (maybe_baggage.if_error() &&
+               test_case.expected_baggage.if_error()) {
+      CHECK(maybe_baggage.error().code ==
+            test_case.expected_baggage.error().code);
+    } else {
+      FAIL("mistmatch between what is expected and the result");
```
xlamorlette-datadog: _nit_: typo: `mismatch`

test/test_trace_segment.cpp
```diff
@@ -175,8 +175,8 @@ TEST_CASE("TraceSegment finalization of spans") {
...
-        "inject") {
+        "'extract_max_size' propagation error if X-Datadog-Tags oversized on "
+        "extract") {
```
xlamorlette-datadog, line 179: _nit_: merge with previous line (we are no more limited to 80 columns ;-) )

test/test_tracer.cpp
```diff
@@ -1248,6 +1248,48 @@ TEST_TRACER("span extraction") {
...
+      "'extract_max_size' propagation error if tracestate \"dd\" vendor "
+      "value is oversized on extract") {
+    std::unordered_map<std::string, std::string> span_tags;
+    MockLogger logger;
+    CAPTURE(logger.entries);
+    CAPTURE(span_tags);
+
+    // Build a "dd" vendor value that exceeds the 512 limit
+    std::string dd_value = "s:1;p:000000003ade68b1;o:synthetics";
+    for (int i = 0; dd_value.size() <= 1024; ++i) {
```
xlamorlette-datadog: _opt_: `513` should be enough instead of `1024` which may be misleading

xlamorlette-datadog (approved):

src/datadog/w3c_propagation.cpp
zacharycmontoya: That's a good idea but I'll leave this as-is for now

test/test_trace_segment.cpp
zacharycmontoya: The formatter did this so who am I disagree 🤷🏼 I'll leave this as-is

test/test_baggage.cpp
zacharycmontoya: Fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

test/test_tracer.cpp
zacharycmontoya: Fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

test/test_trace_segment.cpp
zacharycmontoya: Fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

src/datadog/extraction_util.cpp
zacharycmontoya: fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

src/datadog/extraction_util.cpp
zacharycmontoya: fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

src/datadog/baggage.cpp
zacharycmontoya: fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

src/datadog/baggage.cpp
zacharycmontoya: fixed in d4c0de6511f48083d1435b60ec03422cbddf053a

PR #317, opened 2026-05-06
Bump version following delivery of v2.1.0

xlamorlette-datadog:
Following the [delivery of v2.1.0](https://github.com/DataDog/dd-trace-cpp/releases/tag/v2.1.0), bump the version.

### DIFF
--- a/src/datadog/version.cpp
+++ b/src/datadog/version.cpp
@@ -2,7 +2,7 @@
 
 namespace datadog::tracing {
 
-#define DD_TRACE_VERSION "v2.1.0"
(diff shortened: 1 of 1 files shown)


pawelchcki (approved):

zacharycmontoya (approved):
LGTM. And if we change any small behaviors, we can always bump the minor version

xlamorlette-datadog:
@zacharycmontoya
> And if we change any small behaviors, we can always bump the minor version

Yes, I think that we should bump the minor version in the first PR that justifies it, but this is often overlooked ;-) More realistically, this is done when preparing the following release.

PR #324, opened 2026-06-05
feat: Publish OTel process context

ivoanjo:
# Description

This PR adds support for publishing the OpenTelemetry Process Context (aka OTEP 4719 -- <https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4719-process-ctx.md>).

This mechanism is very similar to Datadog's existing "process discovery"/"tracer-info" mechanism, so the actual change in `trace.cpp` is very minimal (this is by design :p ).

This data will be read by OTel-compliant tools, of which the eBPF Profiler (which Datadog will also support as part of our Continuous Profiler product) is the first one.

# Motivation

We're adding support for OTEP 4719 to all Datadog SDKs, so dd-trace-cpp is actually one of the last ones to adopt this.

See <https://docs.google.com/document/d/1IwjjVJzEChcFPcnVV2N5Kkjg-4_Q4v4Q3ojpxntbdvY/edit?pli=1&tab=t.s7dfru3a9hnl> (Datadog-only link) for an up-to-date list of all SDKs.

# Additional Notes

Jira ticket: [PROF-14789] 

The `otel_process_ctx.cpp` and `otel_process_ctx.h` come from the reference implementation we're maintaining upstream with OTel in <https://github.com/open-telemetry/sig-profiling/tree/main/process-context/c-and-cpp>.

They are already in use in dd-trace-java as well.

Review-wise, my suggestion for those would be to focus on correctness -- ideally we want to be able to quickly update to the latest version when needed, which is made easier if we have a minimal delta on upstream.

(I'm the one maintaining this upstream so I can propagate fixes/changes as well so we don't have to diverge).

**How to test the change?**

This change includes test coverage. Also, I've tested this in practice using the `otel_process_ctx_dump.sh` script from the https://github.com/open-telemetry/sig-profiling/ repo.

Here's how that looks:

```
$ sudo ./otel_process_ctx_dump.sh 2394947
Found OTEL context for PID 2394947
Start address: 7c072aace000
00000000  4f 54 45 4c 5f 43 54 58  02 00 00 00 c3 01 00 00  |OTEL_CTX........|
00000010  68 5b 15 20 9a 25 21 00  60 f0 6e 16 c5 5b 00 00  |h[. .%!.`.n..[..|
00000020
Parsed struct:
  otel_process_ctx_signature             : "OTEL_CTX"
  otel_process_ctx_version               : 2
  otel_process_payload_size              : 451
  otel_process_monotonic_published_at_ns : 9330018124913512
  otel_process_payload                   : 0x00005bc5166ef060
Payload dump (451 bytes):
00000000  0a af 02 0a 21 0a 1b 64  65 70 6c 6f 79 6d 65 6e  |....!..deploymen|
00000010  74 2e 65 6e 76 69 72 6f  6e 6d 65 6e 74 2e 6e 61  |t.environment.na|
00000020  6d 65 12 02 0a 00 0a 3d  0a 13 73 65 72 76 69 63  |me.....=..servic|
00000030  65 2e 69 6e 73 74 61 6e  63 65 2e 69 64 12 26 0a  |e.instance.id.&.|
00000040  24 38 32 65 39 38 38 38  64 2d 33 36 65 62 2d 34  |$82e9888d-36eb-4|
00000050  39 63 66 2d 61 63 63 38  2d 36 38 32 39 34 39 63  |9cf-acc8-682949c|
00000060  34 39 32 37 33 0a 20 0a  0c 73 65 72 76 69 63 65  |49273. ..service|
00000070  2e 6e 61 6d 65 12 10 0a  0e 63 70 70 2d 65 78 70  |.name....cpp-exp|
00000080  65 72 69 6d 65 6e 74 0a  15 0a 0f 73 65 72 76 69  |eriment....servi|
00000090  63 65 2e 76 65 72 73 69  6f 6e 12 02 0a 00 0a 1f  |ce.version......|
000000a0  0a 16 74 65 6c 65 6d 65  74 72 79 2e 73 64 6b 2e  |..telemetry.sdk.|
000000b0  6c 61 6e 67 75 61 67 65  12 05 0a 03 63 70 70 0a  |language....cpp.|
000000c0  21 0a 15 74 65 6c 65 6d  65 74 72 79 2e 73 64 6b  |!..telemetry.sdk|
000000d0  2e 76 65 72 73 69 6f 6e  12 08 0a 06 76 32 2e 31  |.version....v2.1|
000000e0  2e 31 0a 24 0a 12 74 65  6c 65 6d 65 74 72 79 2e  |.1.$..telemetry.|
000000f0  73 64 6b 2e 6e 61 6d 65  12 0e 0a 0c 64 64 2d 74  |sdk.name....dd-t|
00000100  72 61 63 65 2d 63 70 70  0a 0f 0a 09 68 6f 73 74  |race-cpp....host|
00000110  2e 6e 61 6d 65 12 02 0a  00 0a 17 0a 0c 63 6f 6e  |.name........con|
00000120  74 61 69 6e 65 72 2e 69  64 12 07 0a 05 31 32 39  |tainer.id....129|
00000130  37 30 12 8e 01 0a 14 64  61 74 61 64 6f 67 2e 70  |70.....datadog.p|
00000140  72 6f 63 65 73 73 5f 74  61 67 73 12 76 0a 74 65  |rocess_tags.v.te|
00000150  6e 74 72 79 70 6f 69 6e  74 2e 62 61 73 65 64 69  |ntrypoint.basedi|
00000160  72 3a 62 75 69 6c 64 2c  65 6e 74 72 79 70 6f 69  |r:build,entrypoi|
00000170  6e 74 2e 77 6f 72 6b 64  69 72 3a 63 70 70 2d 65  |nt.workdir:cpp-e|
00000180  78 70 65 72 69 6d 65 6e  74 2c 65 6e 74 72 79 70  |xperiment,entryp|
00000190  6f 69 6e 74 2e 74 79 70  65 3a 65 78 65 63 75 74  |oint.type:execut|
000001a0  61 62 6c 65 2c 65 6e 74  72 79 70 6f 69 6e 74 2e  |able,entrypoint.|
000001b0  6e 61 6d 65 3a 63 70 70  2d 65 78 70 65 72 69 6d  |name:cpp-experim|
000001c0  65 6e 74                                          |ent|
000001c3
Protobuf decode:
resource {
  attributes {
    key: "deployment.environment.name"
    value {
      string_value: ""
    }
  }
  attributes {
    key: "service.instance.id"
    value {
      string_value: "82e9888d-36eb-49cf-acc8-682949c49273"
    }
  }
  attributes {
    key: "service.name"
    value {
      string_value: "cpp-experiment"
    }
  }
  attributes {
    key: "service.version"
    value {
      string_value: ""
    }
  }
  attributes {
    key: "telemetry.sdk.language"
    value {
      string_value: "cpp"
    }
  }
  attributes {
    key: "telemetry.sdk.version"
    value {
      string_value: "v2.1.1"
    }
  }
  attributes {
    key: "telemetry.sdk.name"
    value {
      string_value: "dd-trace-cpp"
    }
  }
  attributes {
    key: "host.name"
    value {
      string_value: ""
    }
  }
  attributes {
    key: "container.id"
    value {
      string_value: "12970"
    }
  }
}
extra_attributes {
  key: "datadog.process_tags"
  value {
    string_value: "entrypoint.basedir:build,entrypoint.workdir:cpp-experiment,entrypoint.type:executable,entrypoint.name:cpp-experiment"
  }
}
```

[PROF-14789]: https://datadoghq.atlassian.net/browse/PROF-14789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/datadog/tracer.cpp
```diff
@@ -120,6 +121,10 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
   store_config(process_tags);
 }
 
+Tracer::~Tracer() {
+  otel_process_ctx_drop_current();
+}
```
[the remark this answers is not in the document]
ivoanjo: This is by design -- the context is being treated as global so if multiple tracers exist, we're happy with last-writer-wins for the time being.

src/datadog/otel_process_ctx.cpp
```diff
@@ -0,0 +1,898 @@
...
+    return data_out->deployment_environment_name != NULL &&
+           data_out->service_instance_id != NULL &&
+           data_out->service_name != NULL &&
+           data_out->service_version != NULL &&
+           data_out->telemetry_sdk_language != NULL &&
+           data_out->telemetry_sdk_version != NULL &&
+           data_out->telemetry_sdk_name != NULL;
+  }
+
+  void otel_process_ctx_read_data_drop(otel_process_ctx_data data) {
```
xlamorlette-datadog: _nit_: This should be `static` as other internal functions.

include/datadog/tracer.h
```diff
@@ -65,6 +65,9 @@ class Tracer {
   Tracer(const FinalizedTracerConfig& config,
          const std::shared_ptr<const IDGenerator>& generator);
 
+  // Drops all state, including process discovery and process context.
```
xlamorlette-datadog: _nit, pedantic_: use infinitive, as elsewhere in the file

src/datadog/tracer.cpp
```diff
@@ -120,6 +121,8 @@ Tracer::Tracer(const FinalizedTracerConfig& config,
   store_config(process_tags);
 }
 
+Tracer::~Tracer() { otel_process_ctx_drop_current(); }
```
xlamorlette-datadog: _nit, pedantic_: rather split on three lines

include/datadog/tracer.h
```diff
@@ -65,6 +65,9 @@ class Tracer {
...
+  ~Tracer();
```
xlamorlette-datadog, line 72: Declaring a user-defined destructor suppresses the implicit move constructor and move assignment.

Moreover, the implicit copy is now problematic because it will lead to several calls to `otel_process_ctx_drop_current()`. More precisely, the problem is that:

- `otel_process_ctx_publish()` is called from the constructor (via `store_config()`, line 228);
- `otel_process_ctx_drop_current()` is called in then destructor (line 124).

They operate on a process-global singleton (`published_state` in `otel_process_ctx.cpp`). So the destructor's drop is global, and it fires for every Tracer instance destroyed, including copies and temporaries.

This notably breaks the implementation of `nginx-datadog` that uses move and return by value on Tracer.

I suggest adding a move-only RAII guard, held by a `unique_ptr`, so that to transfer the ownership on move. Copy constructor and assignment should be deleted. This is just an idea, maybe a better solution could come from thinking about the whole Otel process context handling.

By the way, `TEST_TRACER("move semantics")`, line 1885 of `test_tracer.cpp` is wrong: when the move constructor doesn't exist, it binds to the copy constructor. This test should be fixed!

src/datadog/tracer.cpp
```diff
@@ -160,9 +163,16 @@ void Tracer::store_config(
 
   metadata_file_ = std::make_unique<InMemoryFile>(std::move(*maybe_file));
 
-  auto defaults = config_manager_->span_defaults();
-
-  std::string container_id = "";
+  const auto defaults = config_manager_->span_defaults();
+  const std::string runtime_id_string = runtime_id_.string();
```
xlamorlette-datadog: _nit_: rather use references

src/datadog/tracer.cpp
```diff
@@ -172,23 +182,57 @@ void Tracer::store_config(
...
+  };
+
+  otel_process_ctx_data otel_data = {};
+  otel_data.deployment_environment_name = service_env.c_str();
+  otel_data.service_instance_id = runtime_id_string.c_str();
+  otel_data.service_name = service_name.c_str();
+  otel_data.service_version = service_version.c_str();
+  otel_data.telemetry_sdk_language = tracer_language.c_str();
+  otel_data.telemetry_sdk_version = tracer_version_value.c_str();
+  otel_data.telemetry_sdk_name = "dd-trace-cpp";
```
xlamorlette-datadog: _nit, opt_: maybe factorize with what is in `src/datadog/version.cpp` by adding a `tracer_library_name` constant.

test/test_otel_process_ctx.cpp
```diff
@@ -0,0 +1,107 @@
...
+  config.collector = std::make_shared<MockCollector>();
+  config.logger = std::make_shared<MockLogger>();
+
+  auto finalized = finalize_config(config);
+  REQUIRE(finalized);
+
+  {
+    Tracer tracer{*finalized};
+
+    auto read_result = otel_process_ctx_read();
```
xlamorlette-datadog, line 80: I think we should also test the published context from the Tracer, so that to check the Tracer→context field mapping.

test/test_otel_process_ctx.cpp
```diff
@@ -0,0 +1,107 @@
+#include <datadog/runtime_id.h>
```
xlamorlette-datadog: _nit_: `runtime_id.h`,  `tracer_config.h`, `version.h`, `filesystem`, `map`, `string`, unordered_map` are already indirectly included.

test/test_otel_process_ctx.cpp
```diff
@@ -0,0 +1,107 @@
...
+    const std::map<std::string, std::string> expected_resource = {
+        {"host.name", get_hostname()},
+        {"container.id", expected_container_id},
+    };
+    CHECK(to_map(data.resource_attributes) == expected_resource);
+
+    const std::map<std::string, std::string> expected_extra = {
+        {"datadog.process_tags", join_tags(expected_process_tags)},
+    };
+    CHECK(to_map(data.extra_attributes) == expected_extra);
```
xlamorlette-datadog: Comparing unsorted maps is brittle.

test/test_otel_process_ctx.cpp
xlamorlette-datadog, line 1: To be checked if relevant: Maybe add a test with `TracerConfig.report_hostname = false`. Then `host.name` should be empty.

xlamorlette-datadog (changes requested):
This makes the Tracer no more thread safe. I'm not sure if this is fine or not.
We should investigate deeper about this, and at least publicize clearly this constraint.

src/datadog/tracer.cpp
ivoanjo: 👍 Changed in 4f6916ddc618c9ed35f3e02f021aa2905541e62.

test/test_otel_process_ctx.cpp
ivoanjo: Nice catch -- fixed in afddc6daebde1b6dfa62ee4206a9c4414435030

src/datadog/tracer.cpp
ivoanjo: 👍 Fixed in 8b9a8a39b87b96e619450d6cc0e2dc5f5745687

test/test_otel_process_ctx.cpp
ivoanjo: 👍, fixed in c191436497f5f1aab24da2dc9fecd6d78470db2

test/test_otel_process_ctx.cpp
ivoanjo: I'm not sure I understood this suggestion -- do you mean checking the contents of the "datadog-tracer-info-somethingsomething" file as well, or something else?

(In particular, that code path is already covered I believe in `test_tracer.cpp`?)

test/test_otel_process_ctx.cpp
ivoanjo: In 9d050df8826a599d19f6878b0e4096512ffb0c5 I added a test + ended up going with omitting the key when the hostname was empty.

Before we were still including the empty string. Since we don't have a lot of consumers of this data it's unclear if it makes any difference, but I think it's closer to the intent of "report_hostname" => you get no host.name.

include/datadog/tracer.h
ivoanjo: Fixed in 2984054d6dec233a4f9cab3d5596a9c741f78df

src/datadog/otel_process_ctx.cpp
ivoanjo: Fixed in ab0b053f04cebf2eca9297d46521d95e1c3da86

include/datadog/tracer.h
ivoanjo: Ughhh thanks for spotting this, my C++-fu is not strong enough to have spotted this one.

In c5615244c7dfab8e1c3a1281128dc4e0b8376dc I've added a guard and a mutex that enforces thread-safe last-writer-wins semantics for the process context.

And in 0575cdea1ae78c42054f5eb970a5ee47d52fc6d I've tightened the "move semantics" test to not surprise us again in the future.

ivoanjo (commented):
First of all! Thanks for the through + quick review and big thanks for the patience with me taking such a long time to get back to this. Past few days have been very busy and it took me a bit to address all the feedback

> This makes the Tracer no more thread safe. I'm not sure if this is fine or not.
> We should investigate deeper about this, and at least publicize clearly this constraint.

Good point. On top of the RAII guard, I've added a mutex to ensure thread-safety.

src/datadog/otel_process_ctx_guard.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <memory>
+
+#include "otel_process_ctx.h"
+
+namespace datadog {
```
xlamorlette-datadog: Please use compound namespaces:
```
namespace datadog::tracing {
```

Please check this in all new files (and, optionally, in all modified files).

src/datadog/otel_process_ctx_guard.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <datadog/logger.h>
+
+#include <mutex>
+#include <ostream>
+
+namespace datadog {
+namespace tracing {
+namespace {
+
+std::mutex& otel_ctx_mutex() {
```
xlamorlette-datadog: _nit_: rather use verb for functions: `get_otel_context_mutex()`

src/datadog/otel_process_ctx_guard.cpp
```diff
@@ -0,0 +1,39 @@
...
+  static std::mutex m;
```
xlamorlette-datadog: _opt, clean code_: please avoid one letter variable name (C++ is not Go ;-) )

src/datadog/otel_process_ctx_guard.cpp
```diff
@@ -0,0 +1,39 @@
...
+  return m;
+}
+
+}  // namespace
+
+OtelCtxGuard::~OtelCtxGuard() {
+  std::lock_guard<std::mutex> lock(otel_ctx_mutex());
+  otel_process_ctx_drop_current();
```
xlamorlette-datadog: I think this shouldn't work:

Each `OtelCtxGuard` drop the global context unconditionally on destroy. From an old `Tracer` destroying an old guard would drop the context that a newer, still-alive `Tracer` had since published.

Here is a proposed solution. Please check in detail my proposal and challenge it!
I propose to use a ref-counted token (`OTelContexToken` behind a `weak_ptr` singleton), which only fires when the last guard referencing the token is destroyed (i.e. when no `Tracer` needs the context anymore).

```
namespace datadog::tracing {

namespace {
std::mutex& get_otel_context_mutex() {
  static std::mutex otel_context_mutex;
  return otel_context_mutex;
}

struct OtelContextToken {
  ~OtelContextToken() {
    std::lock_guard<std::mutex> lock(get_otel_context_mutex());
    otel_process_ctx_drop_current();
  }
};

std::shared_ptr<OtelContextToken> get_otel_context_token() {
  // The weak_ptr allows to observe the lifetime of the token without affecting
  // it. This mechanism ensures that the token is destroyed when the last guard
  // is destroyed, allowing the context to be dropped.
  static std::weak_ptr<OtelContextToken> slot;
  std::shared_ptr<OtelContextToken> token = slot.lock();
  if (!token) {
    token = std::make_shared<OtelContextToken>();
    slot = token;
  }
  return token;
}
}  // namespace

std::shared_ptr<OtelContextToken> publish_otel_process_context(
    const otel_process_ctx_data& data, Logger& logger) {
  std::lock_guard<std::mutex> lock(get_otel_context_mutex());
  const otel_process_ctx_result result = otel_process_ctx_publish(&data);
  if (!result.success) {
    logger.log_error([&](std::ostream& log) {
      log << "Failed to publish OpenTelemetry process context: "
          << result.error_message;
    });
    return nullptr;
  }
  return get_otel_context_token();
}

}  // namespace datadog::tracing
```

src/datadog/otel_process_ctx_guard.h
```diff
@@ -0,0 +1,32 @@
...
+//
+// Multi-instance behavior: the OTel context is a per-process singleton.
+// A successful publish replaces any previously-published context, and
+// destroying any guard drops whatever is current (last writer wins style).
+//
+// This global state is serialized via an internal mutex so this class is
+// thread-safe.
+class OtelCtxGuard {
+ public:
+  ~OtelCtxGuard();
```
xlamorlette-datadog: I thinks this shouldn't work, see comment in `.cpp` file.

include/datadog/tracer.h
```diff
@@ -51,6 +52,8 @@ class Tracer {
   // read to determine if the process is instrumented with a tracer and to
   // retrieve relevant tracing information.
   std::shared_ptr<InMemoryFile> metadata_file_;
+  // Owns the published OpenTelemetry process context, if any.
+  std::unique_ptr<OtelCtxGuard> otel_guard_;
```
xlamorlette-datadog: _nit_: please avoid abbreviations (I think 'OTel' is fine since it is very common in this domain specific context, but I would prefer 'context' over 'ctx'); 'otel_guard' is missing the 'context' meaning, so I suggest: `otel_context_guard`.

src/datadog/tracer.cpp
```diff
@@ -160,9 +165,16 @@ void Tracer::store_config(
 
   metadata_file_ = std::make_unique<InMemoryFile>(std::move(*maybe_file));
 
-  auto defaults = config_manager_->span_defaults();
-
-  std::string container_id = "";
+  const auto defaults = config_manager_->span_defaults();
```
xlamorlette-datadog: _nit, clean code_: please avoid usage of `auto`, to improve readability

src/datadog/tracer.cpp
```diff
@@ -172,23 +184,57 @@ void Tracer::store_config(
...
+    "container_id",    [&](auto& buffer) { return msgpack::pack_string(buffer, container_id); }
   );
   // clang-format on
 
   if (!metadata_file_->write_then_seal(buffer)) {
     logger_->log_error("Either failed to write or seal the configuration file");
+    return;
   }
+
+#ifdef __linux__
```
xlamorlette-datadog: _clean code_: This method is becoming too big. Please try to split it.
Quick suggestion (to be checked and cleaned):
```
struct ProcessMetadata;
void write_metadata_file(const ProcessMetadata& metadata);
void publish_otel_context(const ProcessMetadata& metadata);

struct Tracer::ProcessMetadata {
  std::string runtime_id;
  std::string tracer_version;
  std::string tracer_language;
  std::string hostname;
  bool has_hostname;
  std::string service_name;
  std::string service_env;
  std::string service_version;
  std::string process_tags_joined;
  std::string container_id;
};

void Tracer::store_config(
    const std::unordered_map<std::string, std::string>& process_tags) {
  const auto defaults = config_manager_->span_defaults();
  std::string container_id;
  if (auto maybe_container_id = container::get_id()) {
    container_id = maybe_container_id->value;
  }

  ProcessMetadata metadata{runtime_id_.string(),
                           signature_.library_version,
                           std::string(signature_.library_language),
                           hostname_.value_or(""),
                           hostname_.has_value(),
                           defaults->service,
                           defaults->environment,
                           defaults->version,
                           join_tags(process_tags),
                           std::move(container_id)};

  write_metadata_file(metadata);
#ifdef __linux__
  publish_otel_context(metadata);
#endif
}
```

src/datadog/otel_process_ctx_guard.cpp
xlamorlette-datadog, line 1: Please, I think you should try to write a test the key and complex feature handled by this file. Something as: construct two `Tracer`s concurrently, and check that the context stays published while either is alive, and then drops when the last is destroyed.

test/test_otel_process_ctx.cpp
```diff
@@ -0,0 +1,136 @@
...
+std::map<std::string, std::string> to_map(const char** kv) {
+  std::map<std::string, std::string> out;
+  if (kv == nullptr) return out;
+  for (std::size_t i = 0; kv[i] != nullptr && kv[i + 1] != nullptr; i += 2) {
+    out.emplace(kv[i], kv[i + 1]);
+  }
+  return out;
+}
+
+std::map<std::string, std::string> parse_joined_tags(const std::string& s) {
```
xlamorlette-datadog: _nit, clean code_: please avoid one letter name for variables, especially in function signature

test/test_otel_process_ctx.cpp
```diff
@@ -0,0 +1,136 @@
...
+  std::map<std::string, std::string> out;
+  std::istringstream in(s);
+  for (std::string pair; std::getline(in, pair, ',');) {
+    const auto colon = pair.find(':');
+    out.emplace(pair.substr(0, colon), pair.substr(colon + 1));
```
xlamorlette-datadog: What happens if the colon is the last character? Suggested minimal fix:
```
REQUIRE(colon != std::string::npos);
```

test/test_otel_process_ctx.cpp
```diff
@@ -0,0 +1,136 @@
...
+using namespace datadog::tracing;
+
+#define OTEL_CTX_TEST(x) TEST_CASE(x, "[otel_process_ctx]")
+
+namespace {
+
+std::map<std::string, std::string> to_map(const char** kv) {
+  std::map<std::string, std::string> out;
+  if (kv == nullptr) return out;
+  for (std::size_t i = 0; kv[i] != nullptr && kv[i + 1] != nullptr; i += 2) {
```
xlamorlette-datadog: I think we should rather fail in case of malformed array.
I suggest:
```
std::map<std::string, std::string> to_map(const char** key_value_array) {
  std::map<std::string, std::string> out;
  if (key_value_array == nullptr) return out;
  for (std::size_t index = 0; key_value_array[index] != nullptr; i += 2) {
    REQUIRE(key_value_array[index + 1] != nullptr);
    out.emplace(key_value_array[i], key_value_array[index + 1]);
  }
  return out;
}
```

test/test_otel_process_ctx.cpp
xlamorlette-datadog: I think here we don't test the big `msgpack::pack_map(buffer, …)` call in `Tracer::store_config()`, which is simple but which could contain typos in the correspondances between values and their string names. I don't know how easy it is to test memfd's `msgpack`.

xlamorlette-datadog (changes requested):
I think we must be very careful about thread safety.
(See on-going analysis: [Thread Safety (dd-trace-cpp)](https://datadoghq.atlassian.net/wiki/x/H4QOnwE).)

include/datadog/tracer.h
ivoanjo: Fixed in a3eaa6d

src/datadog/otel_process_ctx_guard.cpp
ivoanjo: Fixed in e7b0121

src/datadog/otel_process_ctx_guard.cpp
ivoanjo: Fixed in 363eb3e

src/datadog/otel_process_ctx_guard.h
ivoanjo: Fixed in 58da92f

src/datadog/tracer.cpp
ivoanjo: Fixed in b3ed966

test/test_otel_process_ctx.cpp
ivoanjo: Fixed all in this file in 219568522a0c2b24ac94ea2530777669fb5e6545

test/test_otel_process_ctx.cpp
ivoanjo: Fixed all in this file in 219568522a0c2b24ac94ea2530777669fb5e6545

test/test_otel_process_ctx.cpp
ivoanjo: Fixed all in this file in 219568522a0c2b24ac94ea2530777669fb5e6545

ivoanjo (commented):
I did a pass on all of the small/low-hanging-fruit comments. I'm going to work on the others next.

test/test_otel_process_ctx.cpp
ivoanjo: Added test for this in a6ed13dab6f3b07291c9f13c75e7771b2fe4a83e . I've pulled a test-only msgpack decoder to avoid needing to rely in building our own.

src/datadog/otel_process_ctx_guard.cpp
ivoanjo: ✅ Added such a test in the latest version, thanks for the suggestion

src/datadog/tracer.cpp
ivoanjo: 👍 I've implemented something like this in the latest version

src/datadog/otel_process_ctx_guard.h
ivoanjo: (Latest version based on your earlier suggestion should address this)

src/datadog/otel_process_ctx_guard.cpp
ivoanjo: The latest version implements this idea, with a slight tweak: when there's > 1 tracer active, the runtime-id is omitted, since each tracer has its own and there's no longer one single value used for the whole process.

Let me know your thoughts!

src/datadog/tracer.cpp
```diff
@@ -14,6 +14,7 @@
 #include <algorithm>
 #include <cassert>
 #include <filesystem>
+#include <vector>
```
xlamorlette-datadog: _nit, super opt_: can be omitted (already indirectly included)

src/datadog/otel_process_ctx_registration.cpp
```diff
@@ -0,0 +1,145 @@
...
+      log << "OpenTelemetry process context fields differ between coexisting "
+             "tracers; keeping the previously-published values";
+    });
+  }
+
+  state.runtime_ids.insert(runtime_id);
+
+  const otel_process_ctx_result result = upsert(state);
+  if (!result.success) {
+    logger.log_error([&](std::ostream& log) {
```
xlamorlette-datadog, line 145: `log_error()` runs while `state.mutex` is held.
The logger can be provided be the library user, and it can be slow.
The context publication can fail (for example, if process_tags exceed the 4096-byte encoder limit).
In such a case, it would repeat on every `Tracer` construction.
So we should rather log after releasing the lock.

Suggested fix:
```
#include <datadog/string_view.h>

Optional<StringView> publish_error;

{
  std::lock_guard<std::mutex> lock(state.mutex);

  […]

  const otel_process_ctx_result result = upsert(state);
  if (!result.success) {
    publish_error = result.error_message;
    unregister(state, runtime_id);
  }
}

if (publish_error) {
  logger.log_error([&](std::ostream& log) {
    log << "Failed to publish OpenTelemetry process context: "
        << *publish_error;
  });
  return nullptr;
}
```

What do you think?

src/datadog/otel_process_ctx_registration.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <vector>
+
+#include "otel_process_ctx.h"
+
+namespace datadog::tracing {
+namespace {
+
+// Process-wide OTel context state, including the mutex that guards it.
+// Accessed through the single `get_otel_context_state()` instance.
+struct StaticOtelContextState {
```
xlamorlette-datadog: _nit_: I think I requested to write `Context` instead of `Ctx`, but, actually, this is incoherent with everything else, notably what is copied from the upstream reference implementation, such as `src/datadog/otel_process_ctx.h`. So `StaticOtelContextState` → `StaticOtelCtxtState`. My bad.

xlamorlette-datadog (approved):
Great!
Now, your time to review my work ;-) Please review:
- [The updates I did in the 'Thread Safety (dd-trace-cpp)' documentation]( https://datadoghq.atlassian.net/wiki/spaces/APMINT/history/6963495967/Thread+Safety+dd-trace-cpp).
- The description of [[Envoy] Share a single runtime_id across all worker threads](https://datadoghq.atlassian.net/browse/IDMPL-663).

src/datadog/otel_process_ctx_registration.cpp
ivoanjo: Sounds good! Added in 1780a203d2bfaa7293866fdc9e1f4f6d58aede09 + with a slight tweak to avoid the other log too

src/datadog/otel_process_ctx_registration.cpp
ivoanjo: Applied in 16e36f9a6c6400f67623d1e425deb89235067565 -- I used `StaticOtelCtxState` not `StaticOtelCtxtState` (e.g. CTX vs CTXT) for consistency, LMK if you really wanted that T

src/datadog/tracer.cpp
ivoanjo: Fixed in e792e0cb57106c6b45f9ae9f5c17f7ba0f2a7c22

ivoanjo:
I had to push force to fix the missing commit signatures. Going to fix that merge conflict next...

PR #328, opened 2026-06-12
Capture proxy related environment variables at Curl client construction to avoid crash

xlamorlette-datadog:
# Description

Read the proxy related environment variables when the `Curl` client is created (in `CurlImpl` constructor), and, later, pass the values to `libcurl` explicitly on each request (in `CurlImpl::post()`).

# Motivation

In some (new) tests for the Datadog Nginx module, we get crashes during shutdown.
This is because the tracer's `libcurl` background thread reads proxy environment variables via `getenv()` on every new connection. But when Nginx shuts down, it changes its worker's process title, which rewrites its environment. This leads to a segfault due to a race condition.

# Additional Notes

This mirrors the approach used in `dd-trace-php` (in [ddtrace_comes_minit_proxy_env()](https://github.com/DataDog/dd-trace-php/blob/master/tracer/coms.c#L79-L120)).

I cleaned a bit the surrounding code, notably by extracting `CurlImpl::configure_handle()` from `CurlImpl::post()`.

### DIFF
--- a/src/datadog/curl.cpp
+++ b/src/datadog/curl.cpp
@@ -66,6 +66,22 @@ bool is_unix_socket(const HTTPClient::URL &url) {
          url.scheme == "https+unix";
 }
 
+StringView resolve_proxy(const HTTPClient::URL &url,
+                         const ProxyConfiguration &config) {
+  if (is_unix_socket(url)) {
+    return "";
+  }
+  const Optional<std::string> &scheme_proxy =
+      url.scheme == "https" ? config.https_proxy : config.http_proxy;
+  if (scheme_proxy) {
+    return *scheme_proxy;
+  }
+  if (config.all_proxy) {
+    return *config.all_proxy;
+  }
+  return "";
+}
+
 void throw_on_error(CURLcode result) {
   if (result != CURLE_OK) {
     throw result;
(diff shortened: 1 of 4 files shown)


src/datadog/curl.cpp
```diff
@@ -21,14 +14,80 @@
...
 CurlLibrary libcurl;
 
+struct ProxyConfiguration {
+  Optional<std::string> all_proxy;
+  Optional<std::string> http_proxy;
+  Optional<std::string> https_proxy;
+  Optional<std::string> no_proxy;
+};
+
+Optional<std::string> environment_variable(const char *name) {
```
zacharycmontoya, line 32: How troublesome would it be to call into our environment type to make the `std::getenv` calls? Ideally we limit access to environment variables to as few files as possible

zacharycmontoya (approved):
LGTM with one small refactor comment

src/datadog/curl.cpp
xlamorlette-datadog: Do you mean adding them to `include/datadog/environment.h`? I didn't want to do so because it would publicize these variables in the public documentation of the C++ Tracer, whereas they are more a kind of implicit feature relying on the running environment.

zacharycmontoya: Ah I see what you're saying. Ok then keeping the variables and lookup here makes sense to me. Thanks!

cataphract (approved):

PR #330, opened 2026-06-30
feat(tracing): add span link support

MilanGarnier:
# Description

Adds OpenTelemetry-style span links to dd-trace-cpp.

- **`ExtractedContext struct`** (`include/datadog/extracted_context.h`)
- **`SpanLink` struct** (`include/datadog/span_link.h`): 128-bit `trace_id`, `span_id`, optional `tracestate`, string `attributes`, optional `flags`.
- **msgpack serialization** (`src/datadog/span_link.cpp`): per-link map with the exact field names and omission rules used by dd-trace-go / dd-trace-py / dd-trace-rs (`trace_id`, `trace_id_high` only when non-zero, `span_id`, `attributes` only when non-empty, `tracestate` only when non-empty, `flags` with high bit set when present).
- **`SpanData::span_links`** (`src/datadog/span_data.h/.cpp`): `std::vector<SpanLink>` field; the span encoder emits a `span_links` array only when non-empty (matches `omitempty` behaviour of other tracers).
- **`Span::add_link(const SpanLink&)`** (`include/datadog/span.h`, `src/datadog/span.cpp`): public API to attach a link to a live span.
- **Parametric endpoint** `POST /trace/span/add_link` (`test/system-tests/`): wires up the cross-language system-test client contract.

# Motivation

This feature is supported by all other tracers and is a dependency for future implementations, such as `DD_TRACE_PROPAGATION_BEHAVIOR_EXTRACT` config (planned).

# Additional Notes

- Unit tests in `test/test_span_link.cpp` and `test/test_span.cpp` exercise serialization and the public API end-to-end.
- System tests : parametric tests enabled in https://github.com/DataDog/system-tests/pull/7238
Jira ticket: [APMAPI-1941]

[APMAPI-1941]: https://datadoghq.atlassian.net/browse/APMAPI-1941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


include/datadog/span.h
```diff
@@ -47,13 +47,15 @@
 
 #include "clock.h"
 #include "optional.h"
+#include "span_link.h"
```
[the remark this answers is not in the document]
MilanGarnier: [6c2f121](https://github.com/DataDog/dd-trace-cpp/pull/330/commits/6c2f121762743b78e1a0850ace09982aafa8ed03)

src/datadog/span_data.cpp
```diff
@@ -131,6 +137,18 @@ Expected<void> msgpack_encode(std::string& destination, const SpanData& span) {
        }, "type", [&](auto& destination) {
          return msgpack::pack_string(destination, span.service_type);
        });
+  if (!result) return result;
+
+  if (has_links) {
+    result = msgpack::pack_string(destination, "span_links");
```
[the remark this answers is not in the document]
MilanGarnier: This is true for v0.5 but not applicable here for v0.4. (details in RFC)

include/datadog/span.h
```diff
@@ -166,6 +168,15 @@ class Span {
...
   void set_source(Source);
 
+  // Add a span link to this span, filled from the specified live span.
+  // The linked span's trace ID, span ID, tracestate, and trace flags are
+  // populated automatically. `attrs` is optional user-supplied attributes.
+  void add_link(const Span& linked, const SpanLinkAttributes& attrs = {});
+  // Add a span link to this span, filled from the specified extracted
+  // distributed-tracing context.
+  void add_link(const ExtractedContext& ctx,
+                const SpanLinkAttributes& attrs = {});
```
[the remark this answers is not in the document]
MilanGarnier: changed scope of ExtractedContext to the parametric test server

test/system-tests/request_handler.cpp
```diff
@@ -345,18 +431,41 @@ void RequestHandler::on_extract_headers(const httplib::Request& req,
...
+  const auto upstream_id = span->parent_id().value_or(0);
+  datadog::tracing::ExtractedContext ctx;
+  ctx.trace_id = span->trace_id();
+  ctx.span_id = upstream_id;
+  for (const auto& hdr : *http_headers) {
+    if (hdr.size() != 2) continue;
+    const auto name = hdr[0].get<std::string>();
+    if (name == "tracestate") {
+      ctx.tracestate = hdr[1].get<std::string>();
+    } else if (name == "traceparent") {
```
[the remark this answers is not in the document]
MilanGarnier: [5311345](https://github.com/DataDog/dd-trace-cpp/pull/330/commits/53113451cd679673224ab617aa85b2e1ea703ef2)

include/datadog/span_link.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <cstdint>
+#include <string>
+#include <unordered_map>
+
+#include "expected.h"
+#include "optional.h"
+#include "trace_id.h"
+
+namespace datadog {
```
xlamorlette-datadog: _nit_: please rather use nested namespace syntax

src/datadog/span.cpp
```diff
@@ -164,6 +165,40 @@ void Span::set_source(Source source) {
...
+void Span::add_link(const Span& linked, const SpanLinkAttributes& attrs) {
+  SpanLink link;
+  link.trace_id = linked.trace_id();
+  link.span_id = linked.id();
+  link.attributes = attrs;
+
+  // Capture injected headers (tracestate, traceparent flags) into a map.
+  struct : DictWriter {
+    std::unordered_map<std::string, std::string> map;
+    void set(StringView k, StringView v) override {
```
xlamorlette-datadog: _nit, clean code_: please use meaningful variable names (this is not Go ;-) )

src/datadog/span_data.cpp
```diff
@@ -75,7 +75,13 @@ void SpanData::apply_config(const SpanDefaults& defaults,
 
 Expected<void> msgpack_encode(std::string& destination, const SpanData& span) {
   // clang-format off
-  msgpack::pack_map(
+  const bool has_links = !span.span_links.empty();
+
+  // 12 always-present fields, plus span_links when there are any.
+  auto result = msgpack::pack_map(destination, has_links ? 13u : 12u);
```
xlamorlette-datadog: _nit_: could we name these magic numbers?

src/datadog/span_data.cpp
```diff
@@ -131,6 +137,18 @@ Expected<void> msgpack_encode(std::string& destination, const SpanData& span) {
...
+    result = msgpack::pack_string(destination, "span_links");
+    if (!result) return result;
+    result = msgpack::pack_array(
+        destination, span.span_links,
+        [](std::string& destination, const SpanLink& link) {
+          return msgpack_encode(destination, link);
+        });
+    if (!result) return result;
+  }
   // clang-format on
```
xlamorlette-datadog: _opt_: to be checked (I could do it): why do we need this clang-format exclusion?

src/datadog/span_link.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <cstddef>
+#include <string>
+
+#include "msgpack.h"
+
+namespace datadog {
+namespace tracing {
+
+Expected<void> msgpack_encode(std::string& destination, const SpanLink& link) {
```
xlamorlette-datadog, line 10: _nit, clean code_: please try to split this big function

test/system-tests/request_handler.cpp
```diff
@@ -236,6 +237,92 @@ void RequestHandler::on_set_meta(const httplib::Request& req,
   res.status = 200;
 }
 
+void RequestHandler::on_add_link(const httplib::Request& req,
```
xlamorlette-datadog, line 275: _nit, clean code_: please try to split this big function

test/system-tests/request_handler.cpp
```diff
@@ -345,18 +432,46 @@ void RequestHandler::on_extract_headers(const httplib::Request& req,
...
     res.set_content(response_body_fail.dump(), "application/json");
     return;
   }
 
-  const auto response_body = nlohmann::json{
-      {"span_id", span->parent_id().value()},
-  };
+  const auto upstream_id = span->parent_id().value_or(0);
+
+  StoredLinkContext stored;
```
xlamorlette-datadog: _nit, clean code_: please try to extract this additional part to avoid make this function, which is already too big, grow

xlamorlette-datadog (commented):
This is only a first very light and superficial review, I should review in detail in a few days.

src/datadog/span_data.cpp
MilanGarnier: this was already present before and the reason is that ` msgpack::pack_map_suffix(...)` was not properly formatted. I can totally remove the comments and stage the formatted version instead

include/datadog/span_link.h
MilanGarnier: Fixed in f62084f3f3c442a22c5c75c5c9bcbf5546ce09e3

src/datadog/span.cpp
MilanGarnier: Clarification in 9ac8c2b6d56ea5a7b95683a7735e7d789e68571a

src/datadog/span_data.cpp
MilanGarnier: Instead of naming and finding a way to ensure this number doesn't break/has to stay up to date, I instead refactored so that any path does one msgpack call (and the method will therefore know the size) f736ab5fb50c0dab08834cb1a952302dbd1d5c31

src/datadog/span_data.cpp
MilanGarnier: This is no longer present since I refactored the function f736ab5fb50c0dab08834cb1a952302dbd1d5c31

src/datadog/span_link.cpp
MilanGarnier: It's hard to split this function without compromising clarity since it is simply encoding all present span link fields, I refactored it make it clearer
0da2fea56b75e5fc1877257885f8ebadecf7b88a

test/system-tests/request_handler.cpp
MilanGarnier: [64882ca](https://github.com/DataDog/dd-trace-cpp/pull/330/commits/64882ca263c9af301f149739518567fd9e6a905d)

test/system-tests/request_handler.cpp
MilanGarnier: [a1a3484](https://github.com/DataDog/dd-trace-cpp/pull/330/commits/a1a348448c07caa8b0f7b854bb8d029e983a70a7)

include/datadog/span.h
```diff
@@ -166,6 +167,13 @@ class Span {
   // Specifies the product (AppSec, DBM) that created this span.
   void set_source(Source);
 
+  // Add a span link to this span, filled from the specified live span.
+  // The linked span's trace ID, span ID, tracestate, and trace flags are
+  // populated automatically. `attrs` is optional user-supplied attributes.
+  void add_link(const Span& linked, const SpanLinkAttributes& attrs = {});
```
zacharycmontoya: This might be going above and beyond what you intended to do, but I suggest that we mirror the OpenTelemetry C++ SDK's [AddLink](https://github.com/open-telemetry/opentelemetry-cpp/blob/4e20f690352ce6c2793efcb6448e51363a2cad18/api/include/opentelemetry/trace/span.h#L163) as much as possible. To do that, I think we should do the following:
1. Move the SpanLink struct you defined in the public API into an internal API, since OpenTelemetry C++ doesn't expose a SpanLink type (it only shows up in the resulting traces)
2. Make a public SpanContext struct similar to what you had with SpanLink, except without a map of attributes
3. Consolidate to one `add_link` API that takes the SpanContext and a map of attributes, similar to what you have here

src/datadog/span.cpp
```diff
@@ -164,6 +165,24 @@ void Span::set_source(Source source) {
...
 
+void Span::add_link(const Span& linked, const SpanLinkAttributes& attrs) {
+  SpanLink link;
+  link.trace_id = linked.trace_id();
+  link.span_id = linked.id();
+  link.attributes = attrs;
+
+  // Populate tracestate/flags from the linked trace's sampling decision, if
+  // one has already been made. Do not force a decision here
+  if (auto context = linked.trace_segment().w3c_link_context(*linked.data_)) {
```
zacharycmontoya: If you end up taking my suggestion to pass in a context, then we can just copy over the tracestate and trace flags from the input SpanContext. There will be no need to get information from a trace segment. As a result, we could remove the `w3c_link_context` call you added to trace segment

src/datadog/span_data.cpp
```diff
@@ -74,66 +76,86 @@ void SpanData::apply_config(const SpanDefaults& defaults,
...
+          [](std::string& destination, const SpanLink& link) {
+            return msgpack_encode(destination, link);
+          });
+    };
+    return msgpack::pack_map(
+        destination, "service", pack_service, "name", pack_name, "resource",
+        pack_resource, "trace_id", pack_trace_id, "span_id", pack_span_id,
+        "parent_id", pack_parent_id, "start", pack_start, "duration",
+        pack_duration, "error", pack_error, "meta", pack_meta, "metrics",
+        pack_metrics, "type", pack_type, "span_links", pack_span_links);
```
zacharycmontoya, line 157: Is there a reason for splitting the behavior when there are no span links? I imagine it would be okay to run this even if we have an empty array of span links

src/datadog/span_link.cpp
```diff
@@ -0,0 +1,73 @@
...
+           [&](std::string& destination) {
+             msgpack::pack_integer(destination, link.trace_id.low);
+             return Expected<void>{};
+           }},
+          {true, "span_id",
+           [&](std::string& destination) {
+             msgpack::pack_integer(destination, link.span_id);
+             return Expected<void>{};
+           }},
+          {link.trace_id.high != 0, "trace_id_high",
```
zacharycmontoya: By default I think we should always record this, even if the value is 0. Every tracer should be emitting 128-bit trace ID's (low + high) these days. If you take this suggestion, then we should also reorder this to write trace_id and trace_id_high one after the other for improved readability


```suggestion
          {true, "trace_id_high",
```

include/datadog/span_link.h
```diff
@@ -0,0 +1,39 @@
...
+#include <unordered_map>
+
+#include "expected.h"
+#include "optional.h"
+#include "trace_id.h"
+
+namespace datadog::tracing {
+
+// Convenience alias: the map type used for span link user attributes.
+using SpanLinkAttributes = std::unordered_map<std::string, std::string>;
```
zacharycmontoya: @xlamorlette-datadog I want the add_link API to be able to accept a map where the attribute values could be one of many types (see https://github.com/open-telemetry/opentelemetry-cpp/blob/4e20f690352ce6c2793efcb6448e51363a2cad18/api/include/opentelemetry/common/attribute_value.h). Is there any easy way to for us to build that forward compatability right now, or should we just tackle that when we get to the point that we can record attributes in general to have a variety of primitive types (rather than strings)?

test/system-tests/request_handler.cpp
```diff
@@ -236,6 +237,105 @@ void RequestHandler::on_set_meta(const httplib::Request& req,
...
+  if (span == active_spans_.cend()) {
+    const auto msg =
+        "on_add_link: span not found for id " + std::to_string(*span_id);
+    VALIDATION_ERROR(res, msg);
+  }
+
+  // The linked context is either another active span or a previously extracted
+  // propagation context, both keyed by `parent_id`.
+  auto linked_it = active_spans_.find(*parent_id);
+  if (linked_it != active_spans_.cend()) {
```
zacharycmontoya, line 301: Going down the path of mirroring the OpenTelemetry C++ API, we could implement a [GetContext()](https://github.com/open-telemetry/opentelemetry-cpp/blob/4e20f690352ce6c2793efcb6448e51363a2cad18/api/include/opentelemetry/trace/span.h#L251) public API on the Span class that would return a SpanContext object with the pertinent details of that Span, rather than having to do context injection to surface the information

src/datadog/span_link.cpp
zacharycmontoya: It seems like you're strongly enforcing at the unit test level that this not be populated when the trace ID only has 64-bits, so feel free to keep as-is, but I wonder why this strict decision is being made

MilanGarnier: > Every tracer should be emitting 128-bit trace ID's (low + high) these days

That was a thing I was not aware about, I'll make the change then. I just was inferring that it was optional from what I had seen previously

include/datadog/span.h
MilanGarnier: I've done it in [move span link into private types, expose spancontext and spanlinkattributes](https://github.com/DataDog/dd-trace-cpp/pull/330/commits/a035447ce1fec97ef33e3e67dd88b91e0c32c510) 
I didn't think about how the OTel C++ SDK was structured, thank you for the insight !

src/datadog/span_data.cpp
MilanGarnier: In other tracers the field is omitted when empty, I was following the same logic. I'm also not satisfied with how it looks that way, the other option I had was using a magic number for the number of fields and several msg_pack_suffix to avoid splitting the logic.

e.g. Go uses :
```go
spanLinks []SpanLink `msg:"span_links,omitempty"`
```

test/test_span_link.cpp
```diff
@@ -24,16 +24,16 @@ nlohmann::json encode_to_json(const SpanLink& link) {
 
 TEST_SPAN_LINK("minimal link encodes only trace_id and span_id") {
   SpanLink link;
-  link.trace_id = TraceID(0x1122334455667788ULL);  // high == 0
+  link.trace_id = TraceID(0x1122334455667788ULL, 0xBBBBBBBBBBBBBBBBULL);
```
zacharycmontoya: I actually preferred leaving the high-bits as 0 to ensure that we always emit the trace_id_high. Can you add a case where we test the high-bits being 0?

MilanGarnier: You're right, I was already testing this in the test right after anyway. I reverted it, thanks !

include/datadog/span_context.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+// This component defines `SpanContext`, the identifying and propagation state
+// of a span. It can be supplied to `Span::add_link` to associate a span with a
+// span in this or another trace.
+
+#include <cstdint>
```
xlamorlette-datadog, line 7: _nit, opt, pedantic_: inclusion of `<cstdint>`, `<string>` and `"optional.h"` can be omitted because they are already indirectly included

include/datadog/span.h
```diff
@@ -147,30 +149,39 @@ class Span {
...
   // using the current time in the destructor.
   void set_end_time(std::chrono::steady_clock::time_point);
   // Specifies the product (AppSec, DBM) that created this span.
   void set_source(Source);
 
+  // Return this span's identifying and propagation state. This does not make a
+  // sampling decision when one has not already been made.
+  SpanContext context() const;
+
+  // Add a span link to this span using the specified context. `attributes` are
```
xlamorlette-datadog: _nit, opt, pedantic_: rather 'Add a link to this span'?

include/datadog/trace_segment.h
```diff
@@ -29,6 +29,7 @@
 #include <cstddef>
 #include <memory>
 #include <mutex>
+#include <string>
```
xlamorlette-datadog: _nit, opt_: can be omitted because already indirectly included

include/datadog/span_context.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "optional.h"
+#include "trace_id.h"
+
+namespace datadog::tracing {
+
+// The map type used for user-supplied span-link attributes.
+using SpanLinkAttributes = std::unordered_map<std::string, std::string>;
+
+struct SpanContext {
```
xlamorlette-datadog: Given how `SpanContext` is used (it is currently only constructed in `span.cpp`), it seems to me that we should forbid the instantiation of a `SpanContext` without a `trace_id` and a `span_id`. So I suggest to convert `struct SpanContext` to a class with only one constructor taking `trace_id` and `span_id` as parameters.

include/datadog/trace_segment.h
```diff
@@ -108,18 +109,27 @@ class TraceSegment {
   const Optional<std::string>& origin() const;
   Optional<SamplingDecision> sampling_decision() const;
 
+  // Return the W3C tracestate encoding and derived trace flags for `span`,
+  // based on this segment's sampling decision. Unlike `inject`, this never
+  // makes a sampling decision if one hasn't been made yet -- it returns
+  // `nullopt` instead of forcing (and thereby permanently finalizing) one.
+  // Used by `Span::context` without prematurely deciding sampling for the
+  // trace.
+  Optional<std::pair<std::string, std::uint32_t>> w3c_link_context(
```
xlamorlette-datadog: _opt_: May we give a type alias to `Optional<std::pair<std::string, std::uint32_t>>`?

src/datadog/span.cpp
```diff
@@ -164,6 +164,31 @@ void Span::set_source(Source source) {
                                             to_tag(source));
 }
 
+SpanContext Span::context() const {
+  SpanContext context;
+  context.trace_id = trace_id();
+  context.span_id = id();
+
+  if (auto w3c_context = trace_segment_->w3c_link_context(*data_)) {
```
xlamorlette-datadog: _opt, clean code_: if we have a type alias for `Optional<std::pair<std::string, std::uint32_t>>`, use it rather than `auto`

include/datadog/span_context.h
```diff
@@ -0,0 +1,30 @@
...
+#include <string>
+#include <unordered_map>
+
+#include "optional.h"
+#include "trace_id.h"
+
+namespace datadog::tracing {
+
+// The map type used for user-supplied span-link attributes.
+using SpanLinkAttributes = std::unordered_map<std::string, std::string>;
```
xlamorlette-datadog: This should be in `span_link.h`.

src/datadog/span_link.h
```diff
@@ -0,0 +1,22 @@
...
+// This component defines the internal span-link representation and its
+// MessagePack serialization. Applications use `SpanContext` and
+// `Span::add_link` instead.
+
+#include <datadog/expected.h>
+#include <datadog/span_context.h>
+
+namespace datadog::tracing {
+
+struct SpanLink {
```
xlamorlette-datadog: As for `SpanContext`, I suggest converting this `struct` to a `class` to avoid instantiating without a given `trace_id` and `span_id`.

src/datadog/span_link.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+// This component defines the internal span-link representation and its
+// MessagePack serialization. Applications use `SpanContext` and
+// `Span::add_link` instead.
+
+#include <datadog/expected.h>
```
xlamorlette-datadog: _opt, nit_: can be omitted

src/datadog/span.cpp
```diff
@@ -164,6 +164,31 @@ void Span::set_source(Source source) {
...
+void Span::add_link(const SpanContext& context,
+                    const SpanLinkAttributes& attributes) {
+  SpanLink link;
+  link.trace_id = context.trace_id;
+  link.span_id = context.span_id;
+  link.tracestate = context.tracestate;
+  link.flags = context.flags;
+  link.attributes = attributes;
+
+  data_->span_links.push_back(std::move(link));
```
xlamorlette-datadog: _opt_: with an adequate constructor, we could use `emplace_back`.

src/datadog/span_link.h
```diff
@@ -0,0 +1,22 @@
...
+#include <datadog/span_context.h>
+
+namespace datadog::tracing {
+
+struct SpanLink {
+  TraceID trace_id;
```
xlamorlette-datadog: Rather than duplicating the fields of `SpanContext`, could we compose one in `SpanLink`?

src/datadog/span_link.cpp
```diff
@@ -0,0 +1,73 @@
...
+           [&](std::string& destination) {
+             // The high bit marks "flags is present" so a receiver can
+             // distinguish an explicit value of 0 from an omitted field.
+             msgpack::pack_integer(destination,
+                                   std::uint64_t(*link.flags | (1u << 31)));
+             return Expected<void>{};
+           }},
+      }};
+
+  const auto size = std::count_if(fields.begin(), fields.end(),
```
xlamorlette-datadog: _opt, clean code_: please avoid usage of `auto` for better readability and type checking (except for very complex types)
(same below)

src/datadog/span_link.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <functional>
+#include <string>
+#include <tuple>
+
+#include "msgpack.h"
+
+namespace datadog::tracing {
+
+Expected<void> msgpack_encode(std::string& destination, const SpanLink& link) {
+  // array of (is_present, field_name, pack_function)
```
xlamorlette-datadog: _opt, clean code_: we could replace this comment by the definition of an ad-hoc structure (which would clarify the code line 64)

src/datadog/span_link.cpp
```diff
@@ -0,0 +1,73 @@
...
+  const std::array<
```
xlamorlette-datadog: _opt_: I think it could be simpler to have direct sequential calls, as you did in `msgpack_encode()` in `span_data.cpp`, rather than using type erasure and indirection with a function parameter.

src/datadog/span_data.cpp
```diff
@@ -7,6 +7,8 @@
 
 #include <cassert>
 #include <cstddef>
+#include <string>
```
xlamorlette-datadog, line 10: _opt, nit, minor_: this include can be omitted
(I should very soon formalize such convention…)

src/datadog/span_data.cpp
xlamorlette-datadog: _To be checked_:
We may leverage `pack_map(std::string& buffer, std::size_t size)` as follows:
```
Expected<void> result = msgpack::pack_map(destination, span.span_links.empty() ? 12 : 13);
if (!result) return result;

result = msgpack::pack_map_suffix(
    destination, "service", pack_service, "name", pack_name, "resource",
    pack_resource, "trace_id", pack_trace_id, "span_id", pack_span_id,
    "parent_id", pack_parent_id, "start", pack_start, "duration",
    pack_duration, "error", pack_error, "meta", pack_meta, "metrics",
    pack_metrics, "type", pack_type);
if (!result) return result;

if (span.span_links.empty()) return result;

auto pack_span_links = [&](auto& destination) {
  return msgpack::pack_array(
      destination, span.span_links,
      [](std::string& destination, const SpanLink& link) {
        return msgpack_encode(destination, link);
      });
return msgpack::pack_map_suffix(destination, "span_links", pack_span_links);
```

src/datadog/trace_segment.cpp
```diff
@@ -202,6 +202,35 @@ Optional<SamplingDecision> TraceSegment::sampling_decision() const {
...
+  std::vector<std::pair<std::string, std::string>> trace_tags;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (!sampling_decision_) {
+      return nullopt;
+    }
+    sampling_priority = sampling_decision_->priority;
+    trace_tags = trace_tags_;
+
+    auto& local_root_tags = spans_.front()->tags;
```
xlamorlette-datadog, line 217: _nit, clean code_: please prefer full types than `auto` for readability

src/datadog/trace_segment.cpp
```diff
@@ -202,6 +202,35 @@ Optional<SamplingDecision> TraceSegment::sampling_decision() const {
...
+    auto ts_tag_found = std::find_if(
```
xlamorlette-datadog, line 218: _nit, clean code_: please use explicit variable names: `trace_source_tag_found`

src/datadog/trace_segment.cpp
xlamorlette-datadog, line 217: Lines 217-233 duplicates lines 415-455, 480-482. Please try to factorize.

src/datadog/trace_segment.cpp
```diff
@@ -202,6 +202,35 @@ Optional<SamplingDecision> TraceSegment::sampling_decision() const {
...
+        [](const auto& p) { return p.first == tags::internal::trace_source; });
+    if (ts_tag_found != local_root_tags.cend()) {
+      trace_tags.emplace_back(tags::internal::trace_source,
+                              ts_tag_found->second);
+    }
+  }
+
+  return std::make_pair(
+      encode_tracestate(span.span_id, sampling_priority, origin_, trace_tags,
+                        additional_datadog_w3c_tracestate_,
```
xlamorlette-datadog, line 229: Since we read `additional_datadog_w3c_tracestate_` and `additional_w3c_tracestate_` without holding `mutex_`, I think we should try to make these fields `const` (it seems to me they are never modified, which is great).

include/datadog/span_link.h
xlamorlette-datadog: Since `span_link.h` is no more public, I suppose this is no more an issue.

xlamorlette-datadog (commented):
Here are some technical comments.
Please get approval from @zacharycmontoya who understands the functionality way much better than me.

include/datadog/trace_segment.h
MilanGarnier: I went for an alias to `std::pair<std::string, std::uint32_t>`

include/datadog/span_context.h
MilanGarnier: since Span has an `add_link` method which can take `SpanLink` attributes as an argument, it is not possible to move it in a private header. I moved it alongside `Span`

src/datadog/span_link.cpp
MilanGarnier: I went back for the same encoding style as for `span_data.cpp` (one `pack_map_suffix` for always present attributes and one `pack_map_suffix` per optional attribute), it ends up being cleaner (comment below)

src/datadog/span_data.cpp
MilanGarnier: This works implementing this way

src/datadog/trace_segment.cpp
MilanGarnier: I refactored the search in `find_trace_source_tag`
```c
// If `local_root_tags` contains the `tags::internal::trace_source` tag,
// return its value; otherwise return `nullopt`.
Optional<std::string> find_trace_source_tag(
    const std::unordered_map<std::string, std::string>& local_root_tags) {
  const std::unordered_map<std::string, std::string>::const_iterator
      trace_source_tag_found =
          std::find_if(local_root_tags.cbegin(), local_root_tags.cend(),
                       [](const std::pair<const std::string, std::string>& p) {
                         return p.first == tags::internal::trace_source;
                       });
  if (trace_source_tag_found == local_root_tags.cend()) {
    return nullopt;
  }
  return trace_source_tag_found->second;
}
```

```c
const Optional<std::string> trace_source_tag =
        find_trace_source_tag(spans_.front()->tags);
```

include/datadog/span_context.h
zacharycmontoya: All of the updated API changes that are shown in https://github.com/DataDog/dd-trace-cpp/pull/337 look correct. It would be great to pull them into this branch so this PR has the API and the other one simply utilizes the API

test/system-tests/request_handler.cpp
```diff
@@ -236,6 +237,102 @@ void RequestHandler::on_set_meta(const httplib::Request& req,
...
+    // Propagate sampling priority from the linked span's trace.
```
zacharycmontoya: Is this necessary for system-tests to pass? It seems odd that we're choosing to forcefully overriding the sampling priority of the span when we're adding the span link

test/system-tests/request_handler.cpp
```diff
@@ -345,18 +479,19 @@ void RequestHandler::on_extract_headers(const httplib::Request& req,
...
-    };
+    const auto response_body_fail = nlohmann::json{{"span_id", nullptr}};
     res.set_content(response_body_fail.dump(), "application/json");
     return;
   }
 
-  const auto response_body = nlohmann::json{
-      {"span_id", span->parent_id().value()},
-  };
+  const auto upstream_id = span->parent_id().value_or(0);
```
zacharycmontoya: Can we unify `upstream_id` and `handle` into one variable?

include/datadog/span_link.h
zacharycmontoya: This is still an issue because the public span API for creating a span link exposes a `std::unordered_map<std::string, std::string>` as a parameter:
```cpp
void add_link(const SpanContext& context, const SpanLinkAttributes& attributes = {});
```

Is it possible for us to keep this API today and then later overload it with a more permissive map? I know you're on PTO right now so we might want to take the change now and then update this in a follow-up PR

zacharycmontoya (approved):
LGTM. I have left some comments but non are blocking, just nice-to-haves like keeping all the add_link API changes in this PR and utilizing the API in the behavior extract PR. There was one outstanding question I had about the unordered_map parameter we are accepting because in the future it should be more permissive and allow different types for the values, but I think we can address that later.

test/system-tests/request_handler.cpp
MilanGarnier: It's true that it's misleading, this arrived here because of `test_span_link_propagated_sampling_decisions` and I missed it but I'll mark it as missing feature instead and remove this part

test/system-tests/request_handler.cpp
MilanGarnier: Not really, upstream_id represents an optional span id (and can be 0) but we need a unique key for storing contexts

MilanGarnier:
Thanks! I cherry picked most api changes from PROPAGATION_BEHAVIOR_EXTRACT (except a few changes in parametric test server which were tied with propagation behavior changes) and resolved your other comments !

PR #338, opened 2026-07-06
Huge update of the Design documentation

xlamorlette-datadog:
# Description

This is a huge update of the “Datadog C++ Tracer Design” documentation. Notably:

- Add description of major missing features:
  - Metrics
  - Remote Configuration
  - Trace Context Propagation
  - Baggage Propagation
  - Trace/Span sampling
- Solve some inconsistencies with current implementation (which may come from code changes made since the documentation was written).
- Clean digressions about alternatives that were considered, to concentrate on current state.

# Motivation

This documentation remained nearly untouched since its creation in August 2023. It was a bit cluttered with historical digressions, and it lacked the description of some salient features introduced in the meantime.

# Additional Notes

The 'Components Relationships' diagram is lightly fixed in this PR, but it will get a major update in a following PR, to rather become a summarized class diagram. See https://github.com/DataDog/dd-trace-cpp/pull/339.

Jira ticket: [IDMPL-549 C++ Tracer: update and improve architecture documentation](https://datadoghq.atlassian.net/browse/IDMPL-549)

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -217,10 +217,12 @@ specified, then default implementations are used:
 - [class ThreadedEventScheduler : public EventScheduler](../src/datadog/threaded_event_scheduler.h),
   which uses a dedicated thread for executing scheduled events at the correct time.
 
-`DatadogAgent::flush()` is periodically called by the event scheduler. `flush()` uses the HTTP
-client to send a `POST` request to the Datadog Agent's
-[/v0.4/traces](https://github.com/DataDog/datadog-agent/blob/9d57c10a9eeb3916e661d35dbd23c6e36395a99d/pkg/trace/api/version.go#L22)
-endpoint. It's all callback-based.
+`DatadogAgent::flush()` is periodically called by `EventScheduler`. `flush()` uses the HTTP
+client to send a `POST` request to the Datadog Agent's `/v0.4/traces` endpoint. It's all
+callback-based.
+
+`DatadogAgent` also periodically polls the Agent's `/v0.7/config` endpoint for remote configuration
+updates, via `EventScheduler`. Responses are dispatched to registered `remote_config::Listener`s.
 
 ### HTTPClient
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,9 +1,7 @@
 # Datadog C++ Tracer Design
 
 The primary purpose of this guide is to describe salient features of the Datadog C++ Tracer's
-design. It differs considerably from its [older
-sibling](https://github.com/DataDog/dd-opentracing-cpp) and
-[peers](https://github.com/open-telemetry/opentelemetry-cpp).
+design.
 
 ## Architecture
 
@@ -20,7 +18,7 @@ Each span:
 - contains information about whether an error occurred during the represented operation, including
   an error message, error type, and stack trace,
 - includes an arbitrary name/value mapping of strings, called "tags",
-- includes an arbritrary name/value mapping of numbers, called "metrics",
+- includes an arbitrary name/value mapping of numbers, called "metrics",
 - has a "start time" indicating when the represented operation began,
 - has a "duration", known only once it finishes, indicating how long the represented operation took.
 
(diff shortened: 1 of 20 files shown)


include/datadog/rate.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-// This component provides a `class`, `Rate`, that is a `double` whose value is
-// between zero and one, inclusive.
+// The `Rate` clarr is a `double` whose value is between zero and one,
```
zacharycmontoya:
```suggestion
// The `Rate` class is a `double` whose value is between zero and one,
```

docs/design.md
```diff
@@ -1,99 +1,64 @@
...
 - has a "name" (sometimes called the "operation name"),
-- has a "resource name," which is a description of the thing that the span is about,
+- has a "resource name", which is a description of the thing that the span is about,
 - contains information about whether an error occurred during the represented operation, including
   an error message, error type, and stack trace,
-- includes an arbitrary name/value mapping of strings, called "tags,"
-- has a start time indicating when the represented operation began,
-- has a duration indicating how long the represented operation took to finish.
+- includes an arbitrary name/value mapping of strings, called "tags",
+- includes an arbitrary name/value mapping of numbers, called "metrics",
```
zacharycmontoya: So it is technically true that we have a map of strings to numbers called "metrics" but this is an implementation detail that we should move away from. We should eventually move closer to having a map of string to primitives resembling [span attributes](https://opentelemetry.io/docs/specs/otel/common/#attribute-collections). Maybe no changes are needed in this PR but we should be changing this in the future.

docs/design.md
```diff
@@ -1,99 +1,64 @@
...
-Each span:
+[Span](../include/datadog/span.h) is the component with which users will interact the most. Each
+span:
 
-- has an "ID,"
-- is associated with a "trace ID,"
-- is associated with a "service," which has a "service type," a "version," and an "environment,"
+- has an "ID",
+- is associated with a "trace ID",
+- is associated with a "service", which has a "service type", a "version", and an "environment",
```
zacharycmontoya: I think we can remove the mention of "service type". Practically we have a service and then each span has a type (e.g. "web") but different spans can have different span types, so to say we have a "service type" is not accurate

docs/design.md
```diff
@@ -104,15 +69,15 @@ that accompany handle-to-interface are avoided.
 
 A "trace" is the entire tree of spans having the same trace ID.
 
-Within one process / worker / service, though, typically there is not an entire trace but only part
-of the trace. Let's call the process / worker / service a "tracer."
+Within one process/worker/service, though, typically there is not an entire trace but only part
+of the trace. Let's call the process/worker/service a "tracer".
 
-One portion of a trace that's passing through the tracer is called a "trace segment." A trace
+One portion of a trace that's passing through the tracer is called a "trace segment". A trace
```
zacharycmontoya: The idea of a "trace segment" is a Datadog-specific concept. Should we do any rewording to indicate that this is our own concept and not a concept shared in the OpenTelemetry community?

docs/design.md
```diff
@@ -129,78 +94,114 @@ block
...
+receiving tracer can create spans belonging to the same trace.
 
-`class Collector` is an interface for sending a `TraceSegment`'s spans somewhere once they're all
-done. It's defined in [collector.h](../include/datadog/collector.h).
+There are two directions:
 
-It's just one function: `send()`. More of a callback than an interface.
+- Injection: `Span::inject()` writes the current trace context into a carrier (e.g. outgoing HTTP
+  headers).
+- Extraction: `Tracer::extract_span()` reads trace context from a carrier (e.g. incoming HTTP
```
zacharycmontoya, line 164: nit: Just a subjective note, I would swap the order to mention extract first and then inject

docs/design.md
```diff
@@ -129,78 +94,114 @@ block
...
+multiple styles for both injection and extraction.
 
-A `Collector` can also be shared by several `Tracer`s.
+A trace's sampling decision, once made, is part of what gets propagated: every tracer the trace
+passes through agrees on whether it's kept or dropped, rather than deciding independently.
+
+Propagation can also be suppressed: if tracing is disabled for a trace and no other product depends
+on it continuing, context is not injected at all.
+
+`Tracer` can also propagate `Baggage`, an OpenTelemetry-like key/value store for application data.
```
zacharycmontoya:
```suggestion
`Tracer` can also propagate W3C Baggage, a key/value store for application data that is not inherently part of a distributed trace but is propagated across services alongside a trace.
```

zacharycmontoya (approved):
Thank you for the update! I left some non-blocking comments but the changes LGTM

docs/design.md
xlamorlette-datadog: Thanks a lot. I'll add a comment about this.

docs/design.md
xlamorlette-datadog: Thanks, I will add such a comment.

PR #343, opened 2026-07-23
Place _dd.apm.enabled:0 in every span

cataphract:
# Description

Although the RFC says otherwise (only to place it on the first span in the chunk), this is the safest way to ensure `_dd.apm.enabled` is encountered during processing and was the consensus decision following an incident.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/src/datadog/trace_segment.cpp
+++ b/src/datadog/trace_segment.cpp
@@ -332,11 +332,6 @@ void TraceSegment::span_finished() {
   for (const auto& span_ptr : spans_) {
     SpanData& span = *span_ptr;
     if (!tracing_enabled_) {
-      // RFC seems to only mandate that this be set if the trace is kept.
-      // However, system-tests expect this to always be set.
-      // Add it all the time; can't hurt
-      // Following incident-57980, go beyond the RFC and add it to every span so
-      // that every trace chunk is marked.
       span.numeric_tags[tags::internal::apm_enabled] = 0;
     }
     if (origin_) {
--- a/src/datadog/trace_segment.cpp
+++ b/src/datadog/trace_segment.cpp
@@ -328,16 +328,17 @@ void TraceSegment::span_finished() {
     }
   }
 
-  // RFC seems to only mandate that this be set if the trace is kept.
-  // However, system-tests expect this to always be set.
-  // Add it all the time; can't hurt
-  if (!tracing_enabled_) {
-    local_root.numeric_tags[tags::internal::apm_enabled] = 0;
-  }
-
   // Some tags are repeated on all spans.
   for (const auto& span_ptr : spans_) {
     SpanData& span = *span_ptr;
+    if (!tracing_enabled_) {
+      // RFC seems to only mandate that this be set if the trace is kept.
+      // However, system-tests expect this to always be set.
+      // Add it all the time; can't hurt
+      // Following incident-57980, go beyond the RFC and add it to every span so
+      // that every trace chunk is marked.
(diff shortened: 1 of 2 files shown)


src/datadog/trace_segment.cpp
```diff
@@ -328,16 +328,17 @@ void TraceSegment::span_finished() {
...
-  // Add it all the time; can't hurt
-  if (!tracing_enabled_) {
-    local_root.numeric_tags[tags::internal::apm_enabled] = 0;
-  }
-
   // Some tags are repeated on all spans.
   for (const auto& span_ptr : spans_) {
     SpanData& span = *span_ptr;
+    if (!tracing_enabled_) {
+      // RFC seems to only mandate that this be set if the trace is kept.
```
xlamorlette-datadog: _nit_: Please try to remove references to the RFC and the incident in this comment:
- I think it's better, if reasonable, to not refer external documentation in code.
- Since this is an open-source component, we should try to avoid referencing resources non-Datadog people can't access.
- We even miss the link or the name of the RFC ;-)

test/test_tracer.cpp
```diff
@@ -2130,6 +2130,30 @@ TEST_TRACER("APM tracing disabled") {
...
+  SECTION("_dd.apm.enabled is added to every span") {
+    auto finalized_config = finalize_config(config, clock);
+    REQUIRE(finalized_config);
+    Tracer tracer{*finalized_config};
+
+    SpanConfig service_entry_config;
+    service_entry_config.service = "child-service";
+
+    {
+      auto root = tracer.create_span();
```
xlamorlette-datadog: _nit, opt_: for better readability, please avoid using `auto` (unless for very long type names)

(I'm going to add this, and some other little things in https://github.com/DataDog/dd-trace-cpp/blob/main/docs/conventions.md.)

test/test_tracer.cpp
```diff
@@ -2130,6 +2130,30 @@ TEST_TRACER("APM tracing disabled") {
...
+      auto child = root.create_child();
+      auto service_entry = root.create_child(service_entry_config);
+      (void)child;
```
xlamorlette-datadog: _nit_: I think we should rather the C++17 `maybe_unused` attribute (and avoid C-style cast):
```
    Span root = tracer.create_span();
    [[maybe_unused]] Span child = root.create_child();
    [[maybe_unused]] Span service_entry = root.create_child(service_entry_config);
```

Actually, we could even avoid naming unused values returned by `create_child()` (but one could keep the names for readability):
```
    Span root = tracer.create_span();
    root.create_child();
    root.create_child(service_entry_config);
```

test/test_tracer.cpp
```diff
@@ -2130,6 +2130,30 @@ TEST_TRACER("APM tracing disabled") {
   TimePoint current_time = default_clock();
   auto clock = [&current_time]() { return current_time; };
 
+  SECTION("_dd.apm.enabled is added to every span") {
```
xlamorlette-datadog, line 2156: Should we add a test to check that the tag is absent when `tracing_enabled_ == true`?

xlamorlette-datadog (approved):

test/test_tracer.cpp
cataphract: Added

xlamorlette-datadog (approved):

cataphract:
There is flakiness in the failing test (looks like a race).

PR #346, opened 2026-07-30
fixing dependabot PR CI

bm1549:
# Description

Skips Datadog test-report and coverage uploads when `DD_CI_VIS_API_KEY` is unavailable. Builds and tests still run, and JUnit uploads still run after a test failure when the key exists. The key itself is scoped to the upload steps.

# Motivation

GitHub does not expose Actions secrets to Dependabot or fork pull requests, and reusable workflows may be called without inherited secrets. These runs were failing after successful builds and tests because upload steps ran with an empty key.

# Additional Notes

Validated with actionlint and credential-condition checks.

Jira ticket: N/A

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -55,7 +55,7 @@ jobs:
       packages: read
     env:
       BUILD_DIR: .build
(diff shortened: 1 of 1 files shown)


zacharycmontoya (approved):
LGTM, thanks for looking into this!

PR #354, opened 2026-08-14
feat(tracing): add OpenTelemetry tracestate sampling

MilanGarnier:
## Summary

Adds W3C `tracestate` support for OpenTelemetry `ot` sampling state.

- Parses `ot` separately from Datadog and other vendor state.
- Retains valid inherited state, drops malformed `rv` and `th` values, and writes sampling values for eligible local probability decisions.
- Carries W3C state only when it belongs to the selected matching trace.

## Testing

- `cmake --build .build -j`
- `./.build/test/tests`

### DIFF
--- a/include/datadog/sampling_decision.h
+++ b/include/datadog/sampling_decision.h
@@ -39,8 +39,9 @@ struct SamplingDecision {
   // The per-second maximum allowed number of "keeps" configured for the limiter
   // consulted in this decision, if any.
   Optional<double> limiter_max_per_second;
-  // The outcome of the probability comparison before rate limiting, if any.
-  Optional<bool> probability_sampled;
+  // Whether the sample rate alone, before the rate limiter, would keep this
+  // trace.
+  Optional<bool> was_probability_sampled;
   // The provenance of this decision.
   Origin origin;
 };
--- a/include/datadog/sampling_decision.h
+++ b/include/datadog/sampling_decision.h
@@ -39,6 +39,8 @@ struct SamplingDecision {
   // The per-second maximum allowed number of "keeps" configured for the limiter
   // consulted in this decision, if any.
   Optional<double> limiter_max_per_second;
+  // The outcome of the probability comparison before rate limiting, if any.
+  Optional<bool> probability_sampled;
   // The provenance of this decision.
   Origin origin;
 };
--- a/src/datadog/trace_sampler.cpp
+++ b/src/datadog/trace_sampler.cpp
@@ -50,9 +50,10 @@ SamplingDecision TraceSampler::decide(const SpanData& span) {
     decision.mechanism = int(rule.mechanism);
     decision.limiter_max_per_second = limiter_max_per_second_;
     decision.configured_rate = rule.rate;
-    const std::uint64_t threshold = max_id_from_rate(rule.rate);
-    decision.probability_sampled = knuth_hash(span.trace_id.low) <= threshold;
-    if (*decision.probability_sampled) {
+    const std::uint64_t threshold = max_id_from_rate(*decision.configured_rate);
+    decision.was_probability_sampled =
+        knuth_hash(span.trace_id.low) <= threshold;
+    if (*decision.was_probability_sampled) {
       if (rule.bypass_limiter) {
         decision.priority = int(SamplingPriority::USER_KEEP);
         return decision;
@@ -92,8 +93,8 @@ SamplingDecision TraceSampler::decide(const SpanData& span) {
   }
 
   const std::uint64_t threshold = max_id_from_rate(*decision.configured_rate);
-  decision.probability_sampled = knuth_hash(span.trace_id.low) <= threshold;
-  if (*decision.probability_sampled) {
+  decision.was_probability_sampled = knuth_hash(span.trace_id.low) <= threshold;
+  if (*decision.was_probability_sampled) {
     decision.priority = int(SamplingPriority::AUTO_KEEP);
   } else {
     decision.priority = int(SamplingPriority::AUTO_DROP);
--- a/src/datadog/trace_sampler.cpp
+++ b/src/datadog/trace_sampler.cpp
@@ -51,7 +51,8 @@ SamplingDecision TraceSampler::decide(const SpanData& span) {
     decision.limiter_max_per_second = limiter_max_per_second_;
     decision.configured_rate = rule.rate;
     const std::uint64_t threshold = max_id_from_rate(rule.rate);
-    if (knuth_hash(span.trace_id.low) <= threshold) {
+    decision.probability_sampled = knuth_hash(span.trace_id.low) <= threshold;
+    if (*decision.probability_sampled) {
       if (rule.bypass_limiter) {
         decision.priority = int(SamplingPriority::USER_KEEP);
         return decision;
@@ -91,7 +92,8 @@ SamplingDecision TraceSampler::decide(const SpanData& span) {
   }
 
   const std::uint64_t threshold = max_id_from_rate(*decision.configured_rate);
-  if (knuth_hash(span.trace_id.low) <= threshold) {
+  decision.probability_sampled = knuth_hash(span.trace_id.low) <= threshold;
+  if (*decision.probability_sampled) {
     decision.priority = int(SamplingPriority::AUTO_KEEP);
   } else {
     decision.priority = int(SamplingPriority::AUTO_DROP);
--- a/src/datadog/trace_segment.cpp
+++ b/src/datadog/trace_segment.cpp
@@ -162,9 +162,9 @@ Optional<std::string> resolve_otel_tracestate(
   }
 
   if (!decision.mechanism || !decision.configured_rate ||
-      !decision.probability_sampled ||
+      !decision.was_probability_sampled ||
       !is_probability_mechanism(*decision.mechanism) ||
-      (*decision.probability_sampled && decision.priority <= 0)) {
+      (*decision.was_probability_sampled && decision.priority <= 0)) {
     return rewrite_otel_tracestate(raw, extract_otel_random_value(raw),
                                    nullopt);
   }
@@ -176,9 +176,9 @@ Optional<std::string> resolve_otel_tracestate(
   threshold = std::min(threshold, max_value - 1);
 
   std::uint64_t random_value = (~knuth_hash(trace_id.low)) >> 8;
-  if (*decision.probability_sampled && random_value < threshold) {
+  if (*decision.was_probability_sampled && random_value < threshold) {
     random_value = threshold;
-  } else if (!*decision.probability_sampled && random_value >= threshold) {
+  } else if (!*decision.was_probability_sampled && random_value >= threshold) {
     random_value = threshold == 0 ? 0 : threshold - 1;
   }
 
--- a/src/datadog/trace_segment.cpp
+++ b/src/datadog/trace_segment.cpp
@@ -6,7 +6,6 @@
 #include <datadog/injection_options.h>
 #include <datadog/logger.h>
 #include <datadog/optional.h>
-#include <datadog/sampling_mechanism.h>
 #include <datadog/span_defaults.h>
 #include <datadog/telemetry/metrics.h>
 #include <datadog/telemetry/telemetry.h>
@@ -16,7 +15,6 @@
 #include <cassert>
 #include <charconv>
 #include <cmath>
-#include <cstdint>
 #include <string>
 #include <unordered_map>
 #include <utility>
--- a/src/datadog/trace_segment.cpp
+++ b/src/datadog/trace_segment.cpp
@@ -186,7 +186,7 @@ Optional<std::string> resolve_otel_tracestate(
   }
 
   constexpr std::uint64_t max_value = UINT64_C(1) << 56;
-  auto threshold = static_cast<std::uint64_t>(
+  std::uint64_t threshold = static_cast<std::uint64_t>(
       std::round((1.0 - decision.configured_rate->value()) *
                  static_cast<double>(max_value)));
   threshold = std::min(threshold, max_value - 1);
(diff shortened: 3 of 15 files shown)


include/datadog/sampling_decision.h
```diff
@@ -39,6 +39,8 @@ struct SamplingDecision {
   // The per-second maximum allowed number of "keeps" configured for the limiter
   // consulted in this decision, if any.
   Optional<double> limiter_max_per_second;
+  // The outcome of the probability comparison before rate limiting, if any.
+  Optional<bool> probability_sampled;
```
xlamorlette-datadog: _nit, personal preference_: I like starting boolean values with is/does/was, so that the condition code reads better. I suggest, with an amended comment suggestion:
```
// Whether the sample rate alone, before the rate limiter, would keep this trace.
Optional<bool> was_probablity_sampled;
```

src/datadog/trace_sampler.cpp
```diff
@@ -51,7 +51,8 @@ SamplingDecision TraceSampler::decide(const SpanData& span) {
     decision.limiter_max_per_second = limiter_max_per_second_;
     decision.configured_rate = rule.rate;
     const std::uint64_t threshold = max_id_from_rate(rule.rate);
```
xlamorlette-datadog: _opt, nit_: by the way, use `*decision.configured_rate` as below, line 94, to increase the similarity between the two code blocks:
```
const std::uint64_t threshold = max_id_from_rate(*decision.configured_rate);
```

src/datadog/trace_segment.cpp
```diff
@@ -151,6 +153,54 @@ Optional<std::string> format_rate(double rate, Logger& logger) {
   return std::string(begin, end);
 }
 
+bool is_probability_mechanism(int mechanism) {
```
xlamorlette-datadog: _nit_: this is only about `SamplingMechanism`, so I suggest either:
- move it to a new private header, such as `sampling_mechanism_util.h`;
- move it to existing `sampling_util.h` private header.

src/datadog/trace_segment.cpp
```diff
@@ -151,6 +153,54 @@ Optional<std::string> format_rate(double rate, Logger& logger) {
...
+      return false;
+  }
+}
+
+Optional<std::string> resolve_otel_tracestate(
+    TraceID trace_id, const SamplingDecision& decision,
+    const Optional<std::string>& inherited) {
+  const StringView raw = inherited ? StringView(*inherited) : StringView{};
+  if (decision.origin != SamplingDecision::Origin::LOCAL) {
+    return inherited ? sanitize_otel_tracestate(raw) : nullopt;
```
xlamorlette-datadog: _nit, opt_: since we test `inherited` a second time, maybe move `raw` construction below:
```
if (decision.origin != SamplingDecision::Origin::LOCAL) {
  return inherited ? sanitize_otel_tracestate(StringView(*inherited)) : nullopt;
}

if (!decision.mechanism || !decision.configured_rate ||
    !decision.probability_sampled ||
    !is_probability_mechanism(*decision.mechanism) ||
    (*decision.probability_sampled && decision.priority <= 0)) {
  const StringView raw = inherited ? StringView(*inherited) : StringView{};
  return rewrite_otel_tracestate(raw, extract_otel_random_value(raw),
                                 nullopt);
}
```

src/datadog/trace_segment.cpp
```diff
@@ -151,6 +153,54 @@ Optional<std::string> format_rate(double rate, Logger& logger) {
...
+  }
+
+  if (!decision.mechanism || !decision.configured_rate ||
+      !decision.probability_sampled ||
+      !is_probability_mechanism(*decision.mechanism) ||
+      (*decision.probability_sampled && decision.priority <= 0)) {
```
xlamorlette-datadog: _nit_: extract this last condition `(*decision.probability_sampled && decision.priority <= 0)` into a named variable

src/datadog/trace_segment.cpp
```diff
@@ -216,14 +268,14 @@ Optional<SamplingDecision> TraceSegment::sampling_decision() const {
 
 Optional<std::pair<std::string, std::uint32_t>> TraceSegment::w3c_link_context(
```
xlamorlette-datadog: _nit_: by the way, please use `W3CLinkContext` instead of `std::pair<std::string, std::uint32_t>`, matching the declaration in `trace_segment.h` line 121

src/datadog/trace_segment.cpp
xlamorlette-datadog: By the way, looking at the `lock_guard` block in `TraceSegment::w3c_link_context()`, I found an inconsistency: this `spans_` read should be moved in the `lock_guard` block above, as the one line 281.

src/datadog/w3c_propagation.cpp
```diff
@@ -38,6 +38,97 @@ constexpr bool is_hexdiglc(const char c) {
...
+         std::all_of(value.begin(), value.end(), is_lowercase_hexdig);
+}
+
+Optional<std::uint64_t> parse_otel_random_value(StringView value) {
+  if (!is_valid_otel_hex(value, otel_sampling_value_size,
+                         otel_sampling_value_size)) {
+    return nullopt;
+  }
+
+  const Expected<std::uint64_t> parsed = parse_uint64(value, 16);
```
xlamorlette-datadog, line 60: _nit_: lines 63-67 are the same than lines 76-80, they can be moved in `bool is_valid_otel_hex()` (to be changed into something as `Optional<std::uint64_t> parse_otel_value()`).

src/datadog/w3c_propagation.cpp
xlamorlette-datadog: Please fix this name: as demonstrated by your new `is_lowercase_hexdig()` function, this is should be named `is_hexdig()` (without `lc` suffix).

src/datadog/w3c_propagation.cpp
```diff
@@ -38,6 +38,97 @@ constexpr bool is_hexdiglc(const char c) {
...
+  if (parsed.if_error()) {
+    return nullopt;
+  }
+  return *parsed;
+}
+
+template <class Function>
+void for_each_otel_item(StringView raw, Function&& function) {
+  std::size_t begin = 0;
+  while (begin <= raw.size()) {
```
xlamorlette-datadog: Please check if it should be `<` instead of `<=`. For example:
- an empty `raw` invokes `function` with empty arguments;
- a trailing `;` produces a trailing empty item.

If you change to `<`, please consider creating another abstraction to factorize the same itemization loop that is in `parse_w3c_tracestate()`, and maybe also the one in `append_tracestate_entries()`. Note:
- `append_tracestate_entries()` has an additional `member_count < max_w3c_tracestate_members` condition that could over-complexify the abstraction.
- `parse_w3c_tracestate()` and `append_tracestate_entries()` have an additional `trim()` call, but maybe this should be added in `for_each_otel_item()`.

src/datadog/w3c_propagation.cpp
```diff
@@ -38,6 +38,97 @@ constexpr bool is_hexdiglc(const char c) {
...
+                                 : item.substr(separator + 1);
+    function(item, key, value);
+    if (end == StringView::npos) {
+      return;
+    }
+    begin = end + 1;
+  }
+}
+
+bool append_otel_item(std::string& result, StringView item) {
```
xlamorlette-datadog: The returned boolean is never checked.

src/datadog/w3c_propagation.cpp
```diff
@@ -38,6 +38,97 @@ constexpr bool is_hexdiglc(const char c) {
...
+  const Expected<std::uint64_t> parsed = parse_uint64(value, 16);
+  if (parsed.if_error()) {
+    return nullopt;
+  }
+  return *parsed;
+}
+
+Optional<std::uint64_t> parse_otel_threshold(StringView value) {
+  if (!is_valid_otel_hex(value, min_otel_threshold_size,
+                         otel_sampling_value_size)) {
```
xlamorlette-datadog: _nit_: I find it a bit confusing to reuse `otel_sampling_value_size` with a different meaning. I suppose we should rather have a `max_otel_threshold_size`.

src/datadog/w3c_propagation.cpp
```diff
@@ -38,6 +38,97 @@ constexpr bool is_hexdiglc(const char c) {
...
+constexpr std::size_t max_w3c_tracestate_members = 32;
+
+bool is_valid_otel_hex(StringView value, std::size_t minimum_size,
+                       std::size_t maximum_size) {
+  return value.size() >= minimum_size && value.size() <= maximum_size &&
+         std::all_of(value.begin(), value.end(), is_lowercase_hexdig);
+}
+
+Optional<std::uint64_t> parse_otel_random_value(StringView value) {
+  if (!is_valid_otel_hex(value, otel_sampling_value_size,
```
xlamorlette-datadog: _nit_: maybe align `parse_otel_random_value()` and `otel_sampling_value_size()` names to either 'random value' or 'sampling value'

src/datadog/w3c_propagation.cpp
```diff
@@ -272,45 +302,69 @@ void parse_datadog_tracestate(ExtractedData& result, StringView datadog_value) {
...
+                                      : member.substr(separator + 1);
+  if (key == "dd") {
+    if (member_value.size() > max_datadog_tracestate_value_size) {
+      span_tags[tags::internal::propagation_error] = "extract_max_size";
+    } else {
+      parse_datadog_trace_state(result, member_value);
     }
-    return;
+  } else if (key == "ot") {
+    parse_ot_tracestate(result, member_value);
```
xlamorlette-datadog: Should we check the size as for `dd` above (line 327)?

src/datadog/w3c_propagation.cpp
```diff
@@ -272,45 +302,69 @@ void parse_datadog_tracestate(ExtractedData& result, StringView datadog_value) {
...
-// `extract_tracestate` populates the `additional_w3c_tracestate` field of
-// `ExtractedData`, in addition to those populated by
-// `parse_datadog_tracestate`.
-void extract_tracestate(
-    ExtractedData& result, const DictReader& headers,
-    std::unordered_map<std::string, std::string>& span_tags) {
-  const auto maybe_tracestate = headers.lookup("tracestate");
-  if (!maybe_tracestate || maybe_tracestate->empty()) {
-    return;
+// `parse_ot_tracestate` populates `otel_w3c_tracestate` if it has not already
```
xlamorlette-datadog: - Is it intended that, from `parse_w3c_tracestate_member()`, the first `ot=` entry wins whereas the last `dd=` entry wins? If so, I suppose this is worth a comment.
- Maybe add a test with two `ot=` entries.

xlamorlette-datadog (commented):
Mostly _nit_ and _opt_ comments.

src/datadog/trace_segment.cpp
MilanGarnier: a368c3b -> moved it in `sampling_util.h`

include/datadog/sampling_decision.h
MilanGarnier: 62fdc87 : applies the suggestion

src/datadog/trace_sampler.cpp
MilanGarnier: 62fdc87 uses `*decision.configured_rate` instead of `rule.rate` a second time

src/datadog/w3c_propagation.cpp
MilanGarnier: renamed in 21e3a87

src/datadog/w3c_propagation.cpp
MilanGarnier: 2deb5ae

src/datadog/w3c_propagation.cpp
MilanGarnier: You're right it's `<`, not `<=`
2deb5ae: empty `raw`/trailing `;` no longer invoke the callback, and parsing and encoding now share trimmed W3C member iteration with early stopping.

src/datadog/w3c_propagation.cpp
MilanGarnier: 4b9032b : `void append_otel_item();` instead

src/datadog/w3c_propagation.cpp
MilanGarnier: 772c4c1 -> now uses `max_otel_threshold_size`

src/datadog/w3c_propagation.cpp
MilanGarnier: 772c4c1 also contains this rename

src/datadog/w3c_propagation.cpp
MilanGarnier: 0e64820 -> oversized ot values now set `extract_max_size` and are excluded, with regression coverage that preserves the other vendor member.

src/datadog/w3c_propagation.cpp
MilanGarnier: ccccc52
- the first ot member is intentionally retained, (and this behavior is commented)
- It also adds duplicate-ot extraction test.

src/datadog/trace_segment.cpp
MilanGarnier: 3f4e4fc

src/datadog/trace_segment.cpp
MilanGarnier: Addressed in 07b0f8c by constructing the inherited StringView only after the non-local early return.

src/datadog/trace_segment.cpp
MilanGarnier: changed in 0cea919

src/datadog/trace_segment.cpp
MilanGarnier: [6e53618](https://github.com/DataDog/dd-trace-cpp/pull/354/commits/6e5361827fcc8f45a4fdd86823fb24ed59ced221) -> doing this cleanly involves refactoring slightly the way tags are injected, does this change looks good ?

xlamorlette-datadog (approved):

PR #356, opened 2026-08-19
feat(tracing): add DD_TRACE_PROPAGATION_EXTRACT_FIRST

MilanGarnier:
# Description

Adds C++ support for `DD_TRACE_PROPAGATION_EXTRACT_FIRST`. 
Existing documentation for the configuration :  [DD_TRACE_PROPAGATION_EXTRACT_FIRST in Feature Parity Dashboard](https://feature-parity.us1.prod.dog/configurations?viewType=configurations&configKey=356)

# Motivation

Allow extraction to stop after the first successfully extracted trace context, matching the shared tracer configuration.

# Additional Notes

Priority: low

- Generated `supported-configurations.json` with `config-inversion`.
- Unit tests cover configuration precedence, heterogeneous W3C tracestate handling, and fallback after an unsuccessful extractor.

Jira ticket: N/A


xlamorlette-datadog:
For the record: [DD_TRACE_PROPAGATION_EXTRACT_FIRST in Feature Parity Dashboard](https://feature-parity.us1.prod.dog/configurations?viewType=configurations&configKey=356).

src/datadog/tracer.cpp
```diff
@@ -309,7 +310,8 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
     telemetry::counter::increment(metrics::tracer::trace_context::extracted,
                                   {extracted_tag});
 
-    if (!first_style_with_trace_id && data->trace_id.has_value()) {
+    const bool extracted_trace_context = data->trace_id.has_value();
```
xlamorlette-datadog, line 314: I think the `extracted_trace_context` name is misleading:
- It makes me think this variable is a trace context.
- This only checks that a trace id was extracted, not a full context, which, I think, relates to the comment from Codex below.

Actually, because the existing code is not clean, I find it very difficult to reason about this:
- The `data` name does not convey any useful meaning.
- The on-going modification is only about a part of the huge (270 lines!) `Tracer::extract_span()` function.

I suggest adding a dedicated refactoring PR before this one to clean the `Tracer::extract_span()` function:
- At least extract the part to be modified, to clarify its purpose, and improve variable names. This could also allow to unit test easier the new behavior.
- Ideally, split the function in big blocks (skimming the function, it seems to me there are at least 4 different blocks).

src/datadog/tracer.cpp
```diff
@@ -319,6 +321,10 @@ Expected<Span> Tracer::extract_span(const DictReader& reader,
 
     data->headers_examined = audited_reader.entries_found;
     extracted_contexts.emplace(style, std::move(*data));
+
+    if (propagation_extract_first_ && extracted_trace_context) {
+      break;
```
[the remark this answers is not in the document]
xlamorlette-datadog: Indeed, we can see line 375 a test about `parent_id`.
I think we should clarify what is a successful extraction (and I suppose this would deserve a dedicated unit test).
Once again, I would really like to clarify the purpose of this block of code. I would start by extracting  lines 277-343, but that's only a suggestion after a quick analysis, to be challenged.

test/test_tracer.cpp
```diff
@@ -2019,6 +2049,7 @@ TEST_TRACER("heterogeneous extraction") {
     std::vector<PropagationStyle> injection_styles;
     std::unordered_map<std::string, std::string> extracted_headers;
     std::unordered_map<std::string, std::string> expected_injected_headers;
+    bool extract_first = false;
```
xlamorlette-datadog, line 2095: This should be added in the capture block (lines 2121-2126).

xlamorlette-datadog (changes requested):
I find the C++ Tracer code difficult to reason because it is not in a very good state: it contains lot of legacy code, with questionable technical choices (but, unfortunately, many are part of the public interface).
From now on, I think we should try to clean the existing code to avoid deteriorating the situation further.
Especially here: `Tracer::extract_span()` is the biggest function: 273 lines!
Please, I think we should refactor this function (in a dedicated PR) before adding complexity to it.

PR #360, opened 2026-08-25
Fix W3C tracestate parsing

bm1549:
# Description

- Reject the full `dd` member when a semicolon-delimited submember contains optional whitespace. Empty separator-generated submembers remain valid.
- Canonicalize preserved tracestate siblings when `dd` is present, including empty-member removal and whitespace trimming.
- Preserve the position of `ot` state during extraction and reinjection.
- Add extraction and inject-after-extract coverage for these cases.

# Motivation

Malformed `dd` submembers could be partially parsed, and sibling tracestate values could be reinjected with invalid padding. This makes `dd` parsing all-or-nothing without changing valid empty-separator behavior.

# Additional Notes

Validation:
- `git diff --check`
- `clang-format` 14.0.6 dry run
- C++17 syntax-only compilation of `src/datadog/w3c_propagation.cpp`
- Dependency-free extraction and reinjection harness covering empty separators, whitespace-only and padded submembers, outer-member canonicalization, no-`dd` behavior, and `ot` state ordering

The full CMake test build could not complete because a public dependency download stalled.

Jira ticket: [APMS-20291]

[APMS-20291]: https://datadoghq.atlassian.net/browse/APMS-20291

### DIFF
--- a/src/datadog/w3c_propagation.cpp
+++ b/src/datadog/w3c_propagation.cpp
@@ -131,6 +131,18 @@ struct PartiallyParsedTracestate {
   std::string other_entries;
 };
 
+void append_tracestate_member(std::string& destination, StringView member) {
+  member = trim(member);
+  if (member.empty()) {
+    return;
+  }
+
+  if (!destination.empty()) {
+    destination += ',';
+  }
+  append(destination, member);
+}
+
 // Return the separate Datadog-specific and non-Datadog-specific portions of the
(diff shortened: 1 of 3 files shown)


MilanGarnier:
@bm1549 I merged yesterday a [PR](https://github.com/DataDog/dd-trace-cpp/pull/354) which slightly changed the way the tracestate was being parsed and generated, it might be cleaner (and easier) to regenerate a fix from the newer branch.

PR #364, opened 2026-09-02
feat(telemetry): report AppSec product state

e-n-0:
## What does this PR do?

- registers AppSec in `app-started` telemetry with a disabled default state
- preserves caller-supplied AppSec state so integrations remain authoritative
- nests product initialization errors under the corresponding product payload

`dd-trace-cpp` does not own the AppSec lifecycle, so the fallback is disabled. Integrations with AppSec support can populate `telemetry.products` with their actual enabled, version, and error state; that supplied state takes precedence.

## Motivation

C++ tracer `app-started` events currently omit `products.appsec`, unlike the telemetry expected by system-tests and other Datadog tracers. This leaves product registration and startup state unavailable to telemetry consumers.

## Validation

- `cmake --build build -j4`
- `./build/test/tests` (199862 assertions in 135 test cases)
- standalone tracer driver captured `products.appsec` as `{"enabled":false,"version":"v2.2.1"}`
- clangd diagnostics: no errors in changed files

### DIFF
--- a/src/datadog/telemetry/telemetry_impl.cpp
+++ b/src/datadog/telemetry/telemetry_impl.cpp
@@ -617,27 +617,24 @@ std::string Telemetry::app_started_payload() {
     /// is no need to declare it.
     if (product.name == Product::Name::tracing) continue;
 
-    auto p = nlohmann::json{
-        {to_string(product.name),
-         nlohmann::json{
-             {"version", product.version},
-             {"enabled", product.enabled},
-         }},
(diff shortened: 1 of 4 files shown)


e-n-0:
Closing in favor of #365. AppSec lifecycle state is owned by the integration (not dd-trace-cpp), so the core fallback was misleading. The serializer fix remains in #365, and the authoritative NGINX state wiring will be proposed in nginx-datadog.

PR #366, opened 2026-09-02
fix(telemetry): nest product errors in app-started payload

e-n-0:
## What does this PR do?

Serializes telemetry product details directly under `products.<product>` and keeps optional startup errors inside that product object.

The existing object-level `emplace` throws when an integration registers a product, preventing `app-started` telemetry from being sent. This unblocks integrations such as `nginx-datadog` from reporting their authoritative AppSec startup state.

## Validation

- `cmake --build build -j4`
- `./build/test/tests` (199852 assertions in 134 test cases)
- exercised from a local NGINX worker with AppSec enabled and disabled

### DIFF
--- a/test/telemetry/test_telemetry.cpp
+++ b/test/telemetry/test_telemetry.cpp
@@ -330,7 +330,7 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
       }
     }
 
-    SECTION("With AppSec product state") {
+    SECTION("Product error is nested under the product name") {
       client->clear();
 
       Configuration cfg;
@@ -346,8 +346,11 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
                             client, scheduler, *url);
 
       const auto message_batch = nlohmann::json::parse(client->request_body);
-      const auto& appsec =
-          message_batch["payload"][0]["payload"]["products"]["appsec"];
+      const auto& appsec = message_batch.at("payload")
+                               .at(0)
+                               .at("payload")
+                               .at("products")
+                               .at("appsec");
 
       CHECK(appsec ==
             nlohmann::json{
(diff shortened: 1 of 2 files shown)


test/telemetry/test_telemetry.cpp
```diff
@@ -329,6 +329,33 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
...
+                                        12,
+                                        "Error initializing WAF",
+                                        {}});
+
+      auto telemetry =
+          Telemetry::create(*finalize_config(cfg), tracer_signature, logger,
+                            client, scheduler, *url);
+
+      const auto message_batch = nlohmann::json::parse(client->request_body);
+      const auto& appsec =
```
xlamorlette-datadog: _nit_: rather use `at()` to avoid undefined behavior if a key is missing:
```
const auto& appsec = message_batch.at("payload").at(0).at("payload").at("products").at("appsec");
```

test/telemetry/test_telemetry.cpp
```diff
@@ -329,6 +329,33 @@ TELEMETRY_IMPLEMENTATION_TEST("Tracer telemetry lifecycle") {
         }
       }
     }
+
+    SECTION("With AppSec product state") {
```
xlamorlette-datadog: _nit_: the section name should rather describe what is the tested behavior; suggestion: “Product error is nested under the product name”

xlamorlette-datadog (approved):
Neat!

test/telemetry/test_telemetry.cpp
e-n-0: Updated the section name to describe the asserted behavior.

test/telemetry/test_telemetry.cpp
e-n-0: Switched the payload traversal to checked `at()` access.

PR #367, opened 2026-09-03
fix(tracing): avoid unresolved round symbol

e-n-0:
## Summary

- replace `std::round` in OpenTelemetry sampling threshold calculation with equivalent non-negative arithmetic
- preserve threshold behavior at the `0` and `1` sampling-rate boundaries
- prevent libc-free downstream modules from retaining an unresolved `round` symbol

## Root cause

`nginx-datadog` removes its musl libc dependency after linking. The `std::round` call introduced by #354 remained undefined in the module, causing nginx to fail at startup with `undefined symbol: round`.

## Validation

- focused OpenTelemetry sampling test: 52 assertions passed
- full unit suite: 199,862 assertions in 134 test cases passed
- clang-format 14 dry-run passed
- `trace_segment.cpp.o` has no undefined `round` symbol

### DIFF
--- a/src/datadog/trace_segment.cpp
+++ b/src/datadog/trace_segment.cpp
@@ -14,7 +14,6 @@
 #include <array>
 #include <cassert>
 #include <charconv>
-#include <cmath>
 #include <string>
 #include <unordered_map>
 #include <utility>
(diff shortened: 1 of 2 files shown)


e-n-0:
Closing this workaround because the fix now lives at the correct compatibility boundary in DataDog/nginx-datadog#426. The exact x86_64 artifact from nginx-datadog head 4550c26474cae6e63af88de98ebc68a8a26d9ccc loads successfully in nginx:1.31.3-alpine, and its dynamic symbol table has no unresolved round, ceil, or ceilf symbols.

